PR #18, opened 2018-10-16
Add support for ApplicationELB

linefeedse:
Metrics for ApplicationELBs require that you specify multiple dimensions, dimensions that cannot be preconfigured. This PR extends the discovery so that it discovers LoadBalancers (LB) and TargetGroups (TG) in the ApplicationELB Namespace. It then combines TGs with LBs using the ListMetrics interface so that per-TG metrics can be fetched.

Possible future extensions would be to add the per-AZ dimension and possibly cache the metric dimensions to reduce the number of API queries.

TargetGroups and LoadBalancers are both output to the `aws_alb_info` metric. so that they can be joined on `(name)` in PromQL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,16 @@ discovery:
         period: 60
         length: 900
         nilToZero: true
+  - type: "alb"
+    region: eu-west-1
+    searchTags:
+      - Key: kubernetes.io/service-name
+        Value: .*
+    metrics:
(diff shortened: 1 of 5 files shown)


tsupertramp:
Looks all good - I will release it as 0.7.0-alpha.

Wow aws api is so horrible sometimes. Thanks for your efforts to make it easy for other users!

PR #21, opened 2018-10-29
VPN connection metrics, extended statistics and average

### LINKED ISSUE #20: Returned metrics are inconsistent and average crashes the service
I have the following config:
```
discovery:
  - region: us-west-2
    type: 'ec2'
    searchTags:
      - Key: 'aws:elasticmapreduce:instance-group-role'
        Value: 'MASTER'
    metrics:
      - name: CPUUtilization
        statistics:
        - 'Average'
        period: 300
        length: 60
      - name: StatusCheckFailed
        statistics:
        - 'Maximum'
        period: 300
        length: 60
```
My goal is to return all my EMR master nodes and get their CPU Usage as well as if they've had a status check fail.

Whenever I do this a couple of unexpected things happen:
1) YACE crashes with the following message
```
2018/10/22 17:27:38 Parse config..
2018/10/22 17:27:38 Startup completed
2018/10/22 17:27:41 Not implemented statisticsAverage
```
2) When I change the CPUUtilization statistic to Maximum it doesn't crash but sometimes it only returns one or two instances and sometimes it returns none(instead of 11 instances like I expect).

I've tested with both 0.6.1 and 0.7.0-alpha

AndrewChubatiuk:
fixes #20 #22
- Added VPN connection metrics
- Added ExtendedStatistics (percentiles)
- Added Average Statistic

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ YACE is currently in quick iteration mode. Things will probably break in upcomin
   - es - ElasticSearch
   - rds - Relational Database Service
   - s3 - Object Storage
+  - vpn - VPN connection
 * Static metrics support for all cloudwatch metrics
 
 ## Image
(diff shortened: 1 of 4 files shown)


tsupertramp:
Thanks a lot 👍

PR #30, opened 2018-11-22
Allow setting a role to assume per job

nhinds:
Adds a new `role_arn` setting to each job to allow scraping multiple targets with
different roles.

This will allow `yace` to pull data from multiple AWS accounts using cross-account roles, or use a different role to scrape instances in the current account.


nhinds:
Sure - renamed to `roleArn`. Camel case is more consistent, you're right.

PR #36, opened 2018-12-27
Adding delay param in config to fetch delayed metrics

Deepak1100:
it should work for this kind of scenario
https://github.com/ivx/yet-another-cloudwatch-exporter/issues/35

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ discovery:
         - 'Sum'
         period: 60
         length: 900
+        delay: 300
         nilToZero: true
   - type: "alb"
     region: eu-west-1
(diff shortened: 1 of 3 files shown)


tsupertramp:
Nice thanks a lot @Deepak1100 - Should we make delay default 0?

So users can upgrade more easily?

Deepak1100:
@tsupertrampI have tested it. and by default, if you don't specify its print delay as zero.
I think its go Int which is initialized it as 0.

tsupertramp:
@Deepak1100 - Nice will merge and release tomorrow. Thanks!

Issue #46, opened 2019-02-28
Not getting BucketSizeBytes from dynamic S3 metrics (but getting NumberOfObjects OK)


abuchananTW:
Config file:
```
    discovery:
      exportedTagsOnMetrics:
        ec2:
          - Name
        elb:
          - Name
        es:
          - Name
      jobs:
      - region: us-east-2
        type: "ec2"
        searchTags:
          - Key: environment
            Value: ^ci$
        metrics:
          - name: CPUUtilization
            statistics:
            - 'Average'
            period: 60
            length: 600
          - name: DiskReadOps
            statistics:
            - 'Average'
            period: 60
            length: 600
          - name: DiskWriteOps
            statistics:
            - 'Average'
            period: 60
            length: 600
          - name: StatusCheckFailed
            statistics:
            - 'Maximum'
            period: 60
            length: 600
      - region: us-east-2
        type: "elb"
        searchTags:
          - Key: KubernetesCluster
            Value: devtest.k8s.local
        metrics:
          - name: BackendConnectionErrors
            statistics:
            - Sum
            period: 60
            length: 600
          - name: HealthyHostCount
            statistics:
            - Average
            period: 60
            length: 600
          - name: HTTPCode_Backend_5XX
            statistics:
            - Sum
            period: 60
            length: 600
          - name: HTTPCode_ELB_5XX
            statistics:
            - Sum
            period: 60
            length: 600
          - name: Latency
            statistics:
            - Average
            - Maximum
            period: 60
            length: 600
          - name: RequestCount
            statistics:
            - Sum
            period: 60
            length: 600
          - name: SpilloverCount
            statistics:
            - Sum
            period: 60
            length: 600
          - name: SurgeQueueLength
            statistics:
            - Maximum
            - Sum
            period: 60
            length: 600
      - region: us-east-2
        type: "es"
        searchTags:
          - Key: Name
            Value: ci-logs
        metrics:
          - name: CPUUtilization
            statistics:
            - Average
            period: 60
            length: 600
          - name: FreeStorageSpace
            statistics:
            - Minimum
            period: 60
            length: 600
          - name: ClusterIndexWritesBlocked
            statistics:
            - Maximum
            period: 60
            length: 600
          - name: JVMMemoryPressure
            statistics:
            - Maximum
            period: 60
            length: 600
      - region: us-east-2
        type: "s3"
        searchTags:
          - Key: Environment
            Value: ci
        metrics:
[...]

PR #47, opened 2019-02-28
#46 Add ability to specify dimensions on discovery jobs

abuchananTW:
Added the ability to specify dimensions for discovered jobs.

This solved my issue with getting the BucketSizeBytes metric for S3. Note that I don't really know Go, so I didn't try to solve the backwards compatibility issue around the old behavior of adding that one default metric to all S3 queries since I wasn't sure whether you wanted to keep that as a default fallback or not.

### DIFF
--- a/src/abstract.go
+++ b/src/abstract.go
@@ -117,9 +117,9 @@ func scrapeDiscoveryJob(job job, tagsOnMetrics exportedTagsOnMetrics, clientTag 
 
 		wg.Add(len(job.Metrics))
 		go func() {
-			dimensions := getDimensions(resource.Service, resource.ID, clientCloudwatch)
 			for j := range job.Metrics {
 				metric := job.Metrics[j]
+				dimensions := getDimensions(resource.Service, resource.ID, metric.Dimensions, clientCloudwatch)
 				go func() {
 					data := cloudwatchData{
 						ID:               resource.ID,
--- a/src/abstract.go
+++ b/src/abstract.go
@@ -119,7 +119,8 @@ func scrapeDiscoveryJob(job job, tagsOnMetrics exportedTagsOnMetrics, clientTag 
 		go func() {
 			for j := range job.Metrics {
 				metric := job.Metrics[j]
-				dimensions := getDimensions(resource.Service, resource.ID, metric.Dimensions, clientCloudwatch)
+				dimensions := detectDimensionsByService(resource.Service, resource.ID, clientCloudwatch)
+				dimensions = addAdditionalDimensions(dimensions, metric.AdditionalDimensions)
 				go func() {
 					data := cloudwatchData{
 						ID:               resource.ID,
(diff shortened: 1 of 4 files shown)


abuchananTW:
I took a stab at what a reasonable backwards compatible implementation would look like. If a user doesn't specify any dimensions, it adds the current default one per the previous behavior. That would keep it from breaking anyone else's config.

I'm not clear on the ES one, so I left it alone. It may make sense to wrap it similarly to the S3 one (or not).

tsupertramp:
Thanks a lot for your PR @abuchananTW - I will try to review this as fast as possible.

src/aws_cloudwatch.go
```diff
@@ -202,7 +202,7 @@ func queryAvailableDimensions(resource string, namespace *string, clientCloudwat
 	return dimensions
 }
 
-func getDimensions(service *string, resourceArn *string, clientCloudwatch cloudwatchInterface) (dimensions []*cloudwatch.Dimension) {
+func getDimensions(service *string, resourceArn *string, configuredDimensions []dimension, clientCloudwatch cloudwatchInterface) (dimensions []*cloudwatch.Dimension) {
```
tsupertramp: I think we should rename getDimensions to detectDimensionsByService. That way getDimensions is only for auto dimension discovery and does not two things at once.

src/abstract.go
```diff
@@ -117,9 +117,9 @@ func scrapeDiscoveryJob(job job, tagsOnMetrics exportedTagsOnMetrics, clientTag
 
 		wg.Add(len(job.Metrics))
 		go func() {
-			dimensions := getDimensions(resource.Service, resource.ID, clientCloudwatch)
 			for j := range job.Metrics {
 				metric := job.Metrics[j]
+				dimensions := getDimensions(resource.Service, resource.ID, metric.Dimensions, clientCloudwatch)
```
tsupertramp: Lets create a new method addAdditionalDimensions which takes care of manipulating the dimensions if somebody wants to add dimensions.

src/config.go
```diff
@@ -38,13 +38,14 @@ type static struct {
...
-	Name             string   `yaml:"name"`
-	Statistics       []string `yaml:"statistics"`
-	Period           int      `yaml:"period"`
-	Length           int      `yaml:"length"`
-	Delay            int      `yaml:"delay"`
-	NilToZero        bool     `yaml:"nilToZero"`
-	DisableTimestamp bool     `yaml:"disableTimestamp"`
+	Name             string      `yaml:"name"`
+	Statistics       []string    `yaml:"statistics"`
+	Dimensions       []dimension `yaml:"dimensions"`
```
tsupertramp: I would suggest that we rename Dimensions to additionalDimensions in config.go as the user should not be irritated by its name.

abuchananTW:
Thanks for the feedback. I should be able to take a look tomorrow and make some changes.

abuchananTW:
Just made an update that I think addresses each of the requests you made above. Apologies if anything is non-go idiomatic.

tsupertramp:
That looks awesome. I will try to review it the next days and merge it afterwards @abuchananTW - Thanks a lot! - Thanks for typo fixes and prettify of README.md! :))

PR #52, opened 2019-03-11
when the resource string has prefix as targetgroup/ then no need to trim for prefix /loadbalancer     

GeeksWine:
This was done by mistake from my side that instead of trimming "/loadbalancer"  in "else if" condition i did this trimming in "if" condition part which checks for prefix as "/targetgroup". While creating this diff by mistake I added the trimmed logic in "if" condition instead of adding it "else if" condition. For this you have made the relevant changes in
[963942f]( https://github.com/ivx/yet-another-cloudwatch-exporter/commit/963942f6aa98d93c49a9b5e0a7b05fa9fbac2e44)

Removed the unnecessary part.

Really sorry for this silly mistake from my side.


tsupertramp:
Hey i am always happy about contribution!

No need to say sorry! We are working together on a project and mistakes happen!

Really appreciate your PR!

GeeksWine:
Thanks for being kind :))

PR #69, opened 2019-05-26
Add support for retrieving ASG metrics based on tags

wjam:
Whilst evaluating tools to allow the monitoring & alerting of various CloudWatch metrics through Prometheus, I noticed that this tool doesn't support being able to discover metrics from autoscaling groups automatically due to the fact that the resourcegrouptaggingapi doesn't support ASGs.

This change works around the limitation in the resourcegrouptaggingapi by using the ASG API to discover groups as required.


tsupertramp:
Thanks for your PR :)

PR #70, opened 2019-06-17
Fix illegal tag character replacement, add cURL to container, and add EBS example to readme.

cdchris12:
While trying to setup YACE, we found that one of our EBS volumes had a tag which contained an equals sign, "=", which YACE was reporting as illegal and crashing on. I've updated this replacement command to now handle equals signs as expected.

I've also added a bit to the README.md file that shows some basic EBS configuration to get "BurstBalance" stats from those EBS volumes, since EBS isn't covered by the default example configuration.

I've also added the installation of cURL to the container's Dockerfile to assist with troubleshooting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN go build -v -ldflags "-X main.version=$VERSION" -o yace
 
 FROM alpine:latest
 
-RUN apk --no-cache add ca-certificates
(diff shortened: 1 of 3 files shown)


thomaspeitz:
@cdchris12 - Nice thanks a lot for your contribution! - Will release it today.

PR #74, opened 2019-06-26
WIP | adding awsDimension variable in config to fanout additional dimensions

Deepak1100:
awsDimension variable works at job level

```
discovery:
  jobs:
  - type: "elb"
    cloudWatchMaxRetries: 5
    searchTags:
    - Key: kubernetes.io/service-name
      Value: ".*"
    awsDimensions:
    - AvailabilityZone
    region: ap-south-1
    metrics:
      - name: HealthyHostCount
        statistics:
        - 'Minimum'
        period: 60
        length: 60
        delay: 120
      - name: UnHealthyHostCount
        statistics:
        - 'Maximum'
        period: 60
        length: 60
        delay: 120
      - name: RequestCount
        statistics:
        - 'Sum'
        period: 60
        length: 60
        delay: 120
        nilToZero: true
```
for each resource we will make one list metrics api call with filtered dimension. which will contain default dimension(like the loadbalancer name with value) and awsDimension(which will consist of keys) and additionalDimension if specified

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,6 @@ discovery:
         length: 60
   - type: elb
     region: eu-west-1
-    cloudWatchMaxRetries: 0
     awsDimensions:
      - AvailabilityZone
     searchTags:
--- a/README.md
+++ b/README.md
@@ -148,6 +148,9 @@ discovery:
         length: 60
   - type: elb
     region: eu-west-1
+    cloudWatchMaxRetries: 0
+    awsDimensions:
+     - AvailabilityZone
     searchTags:
       - Key: KubernetesCluster
         Value: production-19
--- a/prometheus.go
+++ b/prometheus.go
@@ -29,7 +29,8 @@ type PrometheusCollector struct {
 
 func NewPrometheusCollector(metrics []*PrometheusMetric) *PrometheusCollector {
 	return &PrometheusCollector{
-		metrics: removeDuplicatedMetrics(metrics),
+		//metrics: removeDuplicatedMetrics(metrics),
+		metrics: metrics,
 	}
 }
 
--- a/abstract.go
+++ b/abstract.go
@@ -27,9 +27,10 @@ func scrapeAwsData(config conf) ([]*tagsData, []*cloudwatchData) {
 		go func() {
 			region := &job.Region
 			roleArn := job.RoleArn
+			maxRetries := job.CloudWatchMaxRetries
 
 			clientCloudwatch := cloudwatchInterface{
-				client: createCloudwatchSession(region, roleArn),
+				client: createCloudwatchSession(region, roleArn, maxRetries),
 			}
 
 			clientTag := tagsInterface{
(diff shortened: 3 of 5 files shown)


prometheus.go
```diff
@@ -29,7 +29,8 @@ type PrometheusCollector struct {
 
 func NewPrometheusCollector(metrics []*PrometheusMetric) *PrometheusCollector {
 	return &PrometheusCollector{
-		metrics: removeDuplicatedMetrics(metrics),
```
Deepak1100: removeDuplicatedMetrics this function will remove duplicate metrics based on Prometheus name label
so from below metrics last 2 metrics will be dropped. which is not valid.
so i think we have to think logic of removeDuplicatedMetrics.
```
aws_elb_healthy_host_count_minimum{dimension_AvailabilityZone="ap-southeast-1a",dimension_LoadBalancerName="abc",name="arn:aws:elasticloadbalancing:ap-southeast-1:abc"} 17
aws_elb_healthy_host_count_minimum{dimension_AvailabilityZone="ap-southeast-1b",dimension_LoadBalancerName="abc",name="arn:aws:elasticloadbalancing:ap-southeast-1:abc"} 17
aws_elb_healthy_host_count_minimum{dimension_AvailabilityZone="ap-southeast-1c",dimension_LoadBalancerName="abc",name="arn:aws:elasticloadbalancing:ap-southeast-1:abc"} 16```

and why do we even need that

**Comment by Deepak1100** [comment]
I have added cloudWatchMaxRetries option in case if anyone is facing then rate limit problem.

**Comment by thomaspeitz** [comment]
Hey @Deepak1100 Sorry that this was hanging so long. Could you recreate a new PR without the cloud watch retries on the rebased master?

If it is not important for you that commits are from you in the git history. I would mention you in the changelog and just copy the lines by myself and add them.

Greetings, Thomas

**Comment by Deepak1100** [comment]
@thomaspeitz I can remove max retries but then for some people, it might break without increasing there CW API limits(as each metric will make an additional list metrics call to get dimensions value). and max retries is completely optional.

**Comment by Deepak1100** [comment]
I have removed max retries.

File: README.md [hunk 2]
```diff
@@ -148,6 +148,9 @@ discovery:
         length: 60
   - type: elb
     region: eu-west-1
+    cloudWatchMaxRetries: 0
```
thomaspeitz: Can we remove this as well?

prometheus.go
```diff
@@ -29,7 +29,8 @@ type PrometheusCollector struct {
...
+		//metrics: removeDuplicatedMetrics(metrics),
```
thomaspeitz: I am pretty sure that we still need this. It is used to ensure that if somebody finds a metric two times through different tags for example the exporter does not fall. But i will merge this and find a better logic for this use case.

README.md
Deepak1100: updated

thomaspeitz:
Will review the next days merge and release this. Thanks for tidying up the branch and waiting so long!

Greetings, Thomas

PR #83, opened 2019-07-26
Make cloudwatch and tag semaphore concurrency configurable

### LINKED ISSUE #82: 0.13.5-0.13.7 are slow compared to 0.13.4 and older
I am updating from 0.11.0 and have found the current version of yace to be much slower than previous versions.

With a scaled down version of my configuration, versions 0.13.4 and 0.11.0 have about the same performance, taking 11 seconds to scrape CloudWatch metrics in a single example account.
Versions 0.13.5-alpha, 0.13.6, and 0.13.7 all take somewhere between 26-37 seconds to scrape the same configuration in the same account, usually closer to 37 seconds. That's around 3x slower.

My real configuration scrapes the same metrics across 3 additional accounts, in 3 regions (i.e. it scrapes the same set of metrics in 12 different account/region combinations). Versions 0.13.4 and 0.11.0 take 11-12 seconds to scrape all of these metrics, which makes sense as each account/region has independent API rate limiting so the additional requests to different locations don't really slow the process down.
Version 0.13.7 took 5.5 minutes, approximately 25x slower than version 0.13.4.

I suspect that adding another AWS account to scrape will make this even slower, although adding more accounts did not change the scrape time at all in previous versions.

In the debug logs, it appears that the requests are being sent in many tiny batches, each of which takes about 5 seconds to return. It seems like the change to limit the concurrency of AWS requests is the cause of this extreme slowdown. Would it be okay to add a parameter to disable this, increase the limit, or make the limit only apply per role and region?

It also looks like the change from #29 to reduce the number of listMetrics calls has been reverted, so ALB metrics will make a listMetrics call for every metric*ALB combination, instead of once per ALB. That doesn't seem like the biggest contributor to the slowdown, though it will increase the chances of hitting cloudwatch API limits again.

Here's a somewhat-redacted version of my scrape configuration for a single account, in case it helps. The multi-account multi-region is identical except each job is repeated 12 times, with different `region` and `roleArn` values.
```
discovery:
  exportedTagsOnMetrics:
    elb:
    - Name
    - Tag1
    - Tag2
    - Tag3
    - Tag4
    alb:
    - Name
    - Tag1
    - Tag2
    - Tag3
    - Tag4
    rds:
    - Name
    - Tag1
    - Tag2
    - Tag3
    - Tag4
    ebs:
    - Name
    - Tag1
    - Tag2
    - Tag3
    - Tag4
    - MountPoint
    lambda:
    - Name
    - Tag1
    - Tag2
    - Tag3
    - Tag4
    vpn:
    - Name
    - Tag1
    - Tag2
    - Tag3
    - Tag4
  jobs:


  - type: elb
    region: ap-southeast-2

    searchTags:
    - Key: Tag2
      Value: ^(value1|value2|value3|...)$
    metrics:
    - name: HealthyHostCount
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: SpilloverCount
      statistics: [Sum]
      period: 60
      length: 600
      delay: 600
    - name: SurgeQueueLength
      statistics: [Maximum, Average]
      period: 60
      length: 600
      delay: 600
    - name: UnHealthyHostCount
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
  - type: alb
    region: ap-southeast-2

    searchTags:
    - Key: Tag2
      Value: ^(value1|value2|value3|...)$
    metrics:
    - name: HealthyHostCount
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: UnHealthyHostCount
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: UnHealthyHostCount
      statistics: [Sum]
      period: 60
      length: 600
      delay: 600
  - type: rds
    region: ap-southeast-2

    searchTags:
    - Key: Tag2
      Value: ^(value1|value2|value3|...)$
    metrics:
    - name: BurstBalance
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: CPUUtilization
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: DatabaseConnections
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: FreeableMemory
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: FreeStorageSpace
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: SwapUsage
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: ReadIOPS
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: WriteIOPS
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
  - type: lambda
    region: ap-southeast-2

    searchTags:
    - Key: Tag2
      Value: ^(value1|value2|value3|...)$
    metrics:
    - name: Invocations
      statistics: [Sum]
      period: 60
      length: 600
      delay: 600
    - name: Errors
      statistics: [Sum]
      period: 60
      length: 600
      delay: 600
    - name: Duration
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
    - name: Throttles
      statistics: [Sum]
      period: 60
      length: 600
      delay: 600
  - type: vpn
    region: ap-southeast-2

    searchTags:
    - Key: Tag2
      Value: ^(value1|value2|value3|...)$
    metrics:
    - name: TunnelState
      statistics: [Average]
      period: 60
      length: 600
      delay: 600
```

nhinds:
Should resolve #82 if users pass high enough numbers for the new `-tag-concurrency` and `-cloudwatch-concurrency` flags, e.g. `-tag-concurrency 1000 -cloudwatch-concurrency 1000`

### DIFF
--- a/abstract.go
+++ b/abstract.go
@@ -9,8 +9,8 @@ import (
 )
 
 var (
-	cloudwatchSemaphore = make(chan struct{}, 5)
-	tagSemaphore        = make(chan struct{}, 5)
+	cloudwatchSemaphore chan struct{}
+	tagSemaphore        chan struct{}
 )
 
 func scrapeAwsData(config conf) ([]*tagsData, []*cloudwatchData) {
--- a/main.go
+++ b/main.go
@@ -14,10 +14,12 @@ import (
 var version = "custom-build"
 
 var (
-	addr        = flag.String("listen-address", ":5000", "The address to listen on.")
-	configFile  = flag.String("config.file", "config.yml", "Path to configuration file.")
-	debug       = flag.Bool("debug", false, "Add verbose logging")
-	showVersion = flag.Bool("v", false, "prints current yace version.")
+	addr                  = flag.String("listen-address", ":5000", "The address to listen on.")
+	configFile            = flag.String("config.file", "config.yml", "Path to configuration file.")
+	debug                 = flag.Bool("debug", false, "Add verbose logging")
+	showVersion           = flag.Bool("v", false, "prints current yace version.")
+	cloudwatchConcurrency = flag.Int("cloudwatch-concurrency", 5, "Maximum number of concurrent requests to CloudWatch API")
+	tagConcurrency        = flag.Int("tag-concurrency", 5, "Maximum number of concurrent requests to Resource Tagging API")
 
 	supportedServices = []string{
 		"alb",
@@ -73,6 +75,9 @@ func main() {
 	if err := config.load(configFile); err != nil {
 		log.Fatal("Couldn't read ", *configFile, ":", err)
 	}
+
+	cloudwatchSemaphore = make(chan struct{}, *cloudwatchConcurrency)
+	tagSemaphore = make(chan struct{}, *tagConcurrency)
 
 	log.Println("Startup completed")
 


thomaspeitz:
Thanks a lot!

PR #88, opened 2019-08-17
make sure to get most up-to-date datapoint

EladDolev:
this PR fixes two bugs
1. inconsistent metrics with every scrape
2. broken scraping when `addCloudwatchTimestamp` and `nilToZero` are set, and no metrics return


thomaspeitz:
Thanks a lot! Will review on Friday.

PR #90, opened 2019-08-17
improve Dockerfile to cache module download

EladDolev:
`CMD`->`ENTRYPOINT` change might break it for some users that will need to adapt


thomaspeitz:
Looks good. Thanks a lot. Will merge on friday and release.

I like the change of CMD > Entrypoint even if it breaks (again) the exporter for folks. But lets do these things still we are stable.

EladDolev:
Sweet :-)

PR #92, opened 2019-08-20
Add support for SQS queues

alext:
This adds support for getting metrics from SQS queues. Note that by
default SQS queues aren't tagged at all, and therefore don't appear in
the resourcegroupstaggingapi. It's therefore necessary to tag them with
something before they'll appear in this exporter.

### DIFF
--- a/aws_cloudwatch.go
+++ b/aws_cloudwatch.go
@@ -171,6 +171,8 @@ func getNamespace(service *string) *string {
 		ns = "AWS/ElasticMapReduce"
 	case "asg":
 		ns = "AWS/AutoScaling"
+	case "sqs":
+		ns = "AWS/SQS"
 	default:
 		log.Fatal("Not implemented namespace for cloudwatch metric: " + *service)
 	}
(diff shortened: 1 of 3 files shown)


thomaspeitz:
Awesome thanks. Will release on Friday.

starkers:
Thanks for this stuff.. @alext sorry to nag but could you possibly provide a sample snippet of of how you configured this?

I've got the `aws_sqs_info` metric appearing after adding tags to all of the SQS queues, but for the life of me I cannot get an actual number out of any of the metrics Cloudwatch has.. 


EG:

image
---
yet I'm pretty sure there "should" be something in there..

image


I'm very happy to PR updated docs/example config once I know what it should look like

alext:
Hi @starkers 

I'm afraid I'm not sure I can remember exactly how I configured this when I was looking at this. For me it turned out to be too much of a pain to persuade Celery to add tags to the SQS queues, so I ended up using a different approach (getting metrics directly from SQS using https://github.com/jmal98/sqs-exporter).

I do remember something about the length/period of metrics being an issue - IIRC SQS metrics in CloudWatch only have a resolution of 5 minutes, so I think I had to set `length` to at least 300 for it to work (this is the same issue as described for ELBs here: https://github.com/ivx/yet-another-cloudwatch-exporter#help-my-metrics-are-intermittent).

Sorry I can't be more help...

PR #105, opened 2019-10-03
Making docker image use non-root user

whitlekx:
Couldn't push directly to master with tag like the documents explain, so forked and requesting PR into master.

Changes just ensure the docker container is run as a non root user to avoid potential security issues.


thomaspeitz:
Thanks a lot!

PR #109, opened 2019-10-24
Remove hardcoded StorageType dimension from S3 metric

### LINKED ISSUE #85: Unable to scrape S3 request metrics e.g. AllRequests
S3 [request metrics](https://docs.aws.amazon.com/AmazonS3/latest/dev/cloudwatch-monitoring.html#s3-request-cloudwatch-metrics) don't appear to have a `StorageType` dimension, just `BucketName` and `FilterId`. I can *add* a `FilterId` dimension with the `additionalDimensions` setting, but I can't seem to remove the default `StorageType=AllStorageTypes` dimension from the S3 job. Seems similar to #78, but request metrics require the additional dimension `FilterId`.

```
discovery:
  jobs:
  - type: s3
    region: us-west-2
    metrics:
    - name: AllRequests
      statistics: [Sum]
      period: 60
      length: 6000
      delay: 600
      additionalDimensions:
      - name: FilterId
        value: MyFilterId
```

```
2019/07/29 05:38:41 CLI helper - aws cloudwatch get-metric-statistics --metric-name AllRequests --dimensions Name=BucketName,Value=my-bucket-nameName=StorageType,Value=AllStorageTypesName=FilterId,Value=EventLogs --namespace AWS/S3 --statistics Sum --period 60 --start-time 2019-07-29T03:48:41Z --end-time 2019-07-29T05:28:41Z
```

When running the command from the "CLI helper" debug logs (and adding the missing spaces), there are no results. Removing the `Name=StorageType,Value=AllStorageTypes` dimension does return data.

I see at one point in #47 there was a condition wrapped around the "AllStorageTypes" dimension where it wouldn't be added if someone had specified additional dimensions, but that condition was later removed. Could that condition be added back, or a separate option to remove default dimensions be added to the metric? e.g.

```
    - name: AllRequests
      statistics: [Sum]
      period: 60
      length: 6000
      delay: 600
      additionalDimensions:
      - name: FilterId
        value: MyFilterId
      includeS3StorageTypeMetric: false
```

nhinds:
The "AllStorageTypes" value for the StorageType dimension was only
valid for the NumberOfObjects metric. The BucketSizeBytes metric
required specifying a different StorageType, and request metrics
could not be retrieved at all because they do not have a StorageType
dimension.

Users with "NumberOfObjects" in their configuration will now need to
add "StorageType" under additionalDimensions or awsDimensions.

Fixes #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,9 @@ discovery:
           - Average
         period: 86400
         length: 172800
+        additionalDimensions:
+          - name: StorageType
+            value: AllStorageTypes
       - name: BucketSizeBytes
         statistics:
           - Average
--- a/aws_cloudwatch.go
+++ b/aws_cloudwatch.go
@@ -339,7 +339,6 @@ func detectDimensionsByService(service *string, resourceArn *string, clientCloud
 		dimensions = append(dimensions, buildDimension("ClientId", arnParsed.AccountID))
 	case "s3":
 		dimensions = buildBaseDimension(arnParsed.Resource, "BucketName", "")
-		dimensions = append(dimensions, buildDimension("StorageType", "AllStorageTypes"))
 	case "efs":
 		dimensions = buildBaseDimension(arnParsed.Resource, "FileSystemId", "file-system/")
 	case "ebs":


thomaspeitz:
Thanks @nhinds <3

PR #111, opened 2019-10-24
Bump aws-sdk-go to v1.25.18

deanrock:
We are using [IAM Roles for Service Accounts](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) to grant IAM access to K8s containers.

Support for this is not present in the existing version of the library, since it was only added later: https://github.com/aws/aws-sdk-go/pull/2667


thomaspeitz:
Thanks <3

PR #112, opened 2019-10-25
Migrate away from broken actions/docker and actions/bin/filter

nhinds:
Looks like github removed the `actions/docker/` and `actions/bin/` repositories


thomaspeitz:
Thanks @nhinds <3

PR #115, opened 2019-10-29
Update AWS SDK to v1.25.21

nhinds:
v1.25.18 broke support for using STS without specifying a
region in the AWS configuration, which we do when assuming
a role.


thomaspeitz:
Awesome - Thanks

PR #118, opened 2019-12-05
Adding GetMetricsData support

Deepak1100:
Add GetMetricData Cloudwatch API
Added MaxRetries for Resourcegrouptaggingapi, Autoscalingapi
Added Counter 

```
resourceGroupTaggingAPICounter
autoScalingAPICounter, 
cloudwatchGetMetricStatisticsAPICounter
cloudwatchGetMetricDataAPICounter
```
Removed code for GetmetricStatistics for discovery Job

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,18 +69,17 @@ exportedTagsOnMetrics:
 
 ### Auto-discovery job
 
-| Key                  | Description                                                                                         |
-| -------------------- | --------------------------------------------------------------------------------------------------  |
-| region               | AWS region                                                                                          |
-| type                 | Service name, e.g. "ec2", "s3", etc.                                                                |
-| enableMetricData     | Use Cloudwatch GetMetricData Api instead of getmetricstatistics.                                   |
-| length (Default 120) | How far back to request data for in seconds (only when enableMetricData: true).                     |
-| delay                | If set it will request metrics up until `current_time - delay` (only when enableMetricData: true).  |
-| type                 | Service name, e.g. "ec2", "s3", etc.                                                                |
-| roleArn              | IAM role to assume (optional)                                                                       |
-| searchTags           | List of Key/Value pairs to use for tag filtering (all must match), Value can be a regex.            |
-| metrics              | List of metric definitions                                                                          |
-| additionalDimensions | List of dimensions to return beyond the default list per service                                    |
+| Key                  | Description                                                                                |
+| -------------------- | ------------------------------------------------------------------------------------------ |
+| region               | AWS region                                                                                 |
+| type                 | Service name, e.g. "ec2", "s3", etc.                                                       |
+| length (Default 120) | How far back to request data for in seconds                                                |
+| delay                | If set it will request metrics up until `current_time - d
+| type                 | Service name, e.g. "ec2", "s3", etc.                                                       |
+| roleArn              | IAM role to assume (optional)                                                              |
+| searchTags           | List of Key/Value pairs to use for tag filtering (all must match), Value can be a regex.   |
+| metrics              | List of metric definitions                                                                 |
+| additionalDimensions | List of dimensions to return beyond the default list per service                           |
 
 searchTags example:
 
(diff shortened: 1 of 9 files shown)


arnitolog:
Hi @Deepak1100. I deployed a new exporter with your code. But even with **enableMetricData: true** I see that it still uses getMetricStatistics API calls. I see that in ASW cost explorer, and in the metrics from the exporter itself: 
yace_cloudwatch_getmetricdata_requests_total - all zeros,
yace_cloudwatch_getmetricstatistics_requests_total - still growing
Could you please advise how to switch to getMetricData properly?

thomaspeitz:
Currently in hospital. Will be back soon and check that :) thanks for your contribution!

README.md
```diff
@@ -67,14 +67,18 @@ exportedTagsOnMetrics:
...
-| roleArn              | IAM role to assume (optional)                                                            |
-| searchTags           | List of Key/Value pairs to use for tag filtering (all must match), Value can be a regex. |
-| metrics              | List of metric definitions                                                               |
-| additionalDimensions | List of dimensions to return beyond the default list per service                         |
+| Key                  | Description                                                                                         |
+| -------------------- | --------------------------------------------------------------------------------------------------  |
+| region               | AWS region                                                                                          |
+| type                 | Service name, e.g. "ec2", "s3", etc.                                                                |
+| enableMetricData     | Use Cloudwatch GetMetricData Api insisted of getmetricstatistics.                                   |
```
bheight-Zymergen: I think this line should probably read "Use CloudWatch GetMetricData API instead of GetMetricStatistics"

README.md
```diff
@@ -67,14 +67,18 @@ exportedTagsOnMetrics:
...
+| lenght (Default 120) | How far back to request data for in seconds (only when enableMetricData: true).                     |
```
bheight-Zymergen: typo here, should be "length"

aws_cloudwatch.go
```diff
@@ -26,17 +26,21 @@ type cloudwatchInterface struct {
...
-	Statistics             []string
-	Points                 []*cloudwatch.Datapoint
-	NilToZero              *bool
-	AddCloudwatchTimestamp *bool
-	CustomTags             []tag
-	Tags                   []tag
-	Dimensions             []*cloudwatch.Dimension
-	Region                 *string
+	ID                      *string
+	MetricId                *string
```
bheight-Zymergen: In keeping with struct field naming practices I'd recommend we rename this to `MetricID`

README.md
Deepak1100: fixed

README.md
Deepak1100: fixed

aws_cloudwatch.go
Deepak1100: fixed

Deepak1100:
updated the PR

Deepak1100:
sample config:
```
discovery:
  jobs:
  - type: "elb"
    region: ap-south-1
    length: 60
    enableMetricData: true
    delay: 120
    searchTags:
      - Key: kubernetes.io/service-name
        Value: .*
    metrics:
      - name: HealthyHostCount
        statistics:
        - 'Minimum'
        period: 60
  - type: "alb"
    region: ap-south-1
    length: 60
    delay: 120
    enableMetricData: true
    searchTags:
      - Key: kubernetes.io/ingress-name
        Value: .*
    metrics:
      - name: RequestCount
        statistics:
        - 'Sum'
        period: 60
        nilToZero: true
      - name: Latency
        statistics:
        - 'Average'
        period: 60
        nilToZero: true
      - name: SpilloverCount
        statistics:
        - 'Sum'
        - 'Maximum'
        period: 60
        nilToZero: true
```

bheight-Zymergen:
Verified with my current configuration, modified with the new settings, that these changes appear to be working as expected.

thomaspeitz:
Hey, so back at work again :)

I love this PR. Would like to set it live. Is there any reason why we keep GetMetricStatistics. From my point of view we should not support both but only GetMetricData.

This would reduce complexity. What are your thoughts?

Deepak1100:
the option is given so that the existing config should work for all the people after the upgrade as it is.
they can change config for getmetricdata latter(we can deprecate it few versions later).
plus some people can use both the call in same config for different job
But I don't mind removing GetMetricStatistics  call now as well

thomaspeitz:
Will release it this weekend. Sorry for the long delay.

abstract.go
```diff
@@ -187,6 +195,105 @@ func scrapeDiscoveryJob(job job, tagsOnMetrics exportedTagsOnMetrics, clientTag
...
+		mux.Unlock()
+		metricTags := resource.metricTags(tagsOnMetrics)
+		dimensions := detectDimensionsByService(resource.Service, resource.ID, clientCloudwatch)
+		for _, commonJobDimension := range commonJobDimensions {
+			dimensions = append(dimensions, commonJobDimension)
+		}
+
+		wg.Add(len(job.Metrics))
+		go func() {
+			for j := range job.Metrics {
```
daviddetorres: I think here the semaphore for the Cloudwatch concurrency is missed. Else it can provoque throttling.

It should be: 
```
		go func() {
			cloudwatchSemaphore <- struct{}{}
			defer func() {
				<-cloudwatchSemaphore
			}()
			for j := range job.Metrics {
```

abstract.go
```diff
@@ -187,6 +195,105 @@ func scrapeDiscoveryJob(job job, tagsOnMetrics exportedTagsOnMetrics, clientTag
...
+		}()
+	}
+	wg.Wait()
+	maxMetricCount := 100
+	metricDataLength := len(getMetricDatas)
+	partition := int(math.Ceil(float64(metricDataLength) / float64(maxMetricCount)))
+	wg.Add(partition)
+	for i := 0; i < metricDataLength; i += maxMetricCount {
+		go func(i int) {
+			defer wg.Done()
```
daviddetorres: Also here the semaphore for Cloudwatch concurrency should be added:

``` 
		go func(i int) {
			defer wg.Done()
			cloudwatchSemaphore <- struct{}{}
			defer func() {
				<-cloudwatchSemaphore
			}()
			end := i + maxMetricCount
```

Deepak1100: why we need cloudwatchSemaphore here?
cloud watch get metrics data supports at least 25 calls per second and if I am not wrong cloudwatchSemaphore will restrict it to one API call at a time. which will increase Prometheus scrape time

Deepak1100: added Semaphore for tags

Deepak1100:
@thomaspeitz  Removed code for GetmetricStatistics for discovery Job and
Added MaxRetries for Resourcegrouptaggingapi, Autoscalingapi

thomaspeitz:
Thanks a lot <3 - Will release it this week.

PR #120, opened 2019-12-27
kafka support

eminugurkenar:
* Adds AWS/Kafka support 

We also changed removeDuplicatedMetrics little bit , as it only checks "name" label , for kafka broker metrics , it was removing some valid metrics. 

created combileLabels function , in order to combine all labels for given metric. What do you think?

```
func combineLabels(labels map[string]string) string {
	var combinedLabels string
	keys := make([]string, 0, len(labels))
	for k := range labels {
		keys = append(keys, k)
	}
	sort.Strings(keys)
	for _, k := range keys {
		combinedLabels += promString(k) + promString(labels[k])
	}
	return combinedLabels
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ YACE is currently in quick iteration mode. Things will probably break in upcomin
   * sqs - Simple Queue Service
   * vpn - VPN connection
   * asg - Auto Scaling Group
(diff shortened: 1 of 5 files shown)


thomaspeitz:
Thanks a lot @eminugurkenar :)

PR #121, opened 2019-12-27
add some validations

rabunkosar-dd:
Adding some validations:
- If period is not supplied, AWS complains about the default 0 value so forcing some default non-zero values.
- If the ECS resource returned through the filter is a cluster, the assumed dimensions are not there causing panic in the code. Maybe the filter part is a better place for this but this will make sure those are no-ops for the code.


thomaspeitz:
Thanks for your contribution @rabunkosar-dd

PR #126, opened 2020-01-17
Pagination Fix for getMetricsList

eminugurkenar:
* Fallow pages for getMetricsList
we noticed that getMetricsList function calls `ListMetricsRequest` which does not follow `NextToken` so it misses some metric dimensions.


thomaspeitz:
Hey thanks @eminugurkenar for your contribution - Will release this on this weekend.

PR #132, opened 2020-02-17
Refactor debug logging

j-nix:
Hi @thomaspeitz, a few I think useful additions to logging logic if you wish to implement!

Debug logic now only happens once, and utilises the `log.Debug()`
function, rather than log.Print() with multiple checks on the existance
of *debug.

"Standard" logging introduced, human-readable timestamped coloured output for reading in TTY sessions.

The following flags added for logging:
`-logfile` - write logs to directory, defaults to stdout
`-logformat` - format of log standard/json, defaults to json

ps: ignore commit 371657b within this PR, this is because I based this branch off my other work, but should be good to go standalone! 👍

### DIFF
--- a/aws_cloudwatch.go
+++ b/aws_cloudwatch.go
@@ -83,20 +83,18 @@ func createGetMetricStatisticsInput(dimensions []*cloudwatch.Dimension, namespac
 		ExtendedStatistics: extendedStatistics,
 	}
 
-	if *debug {
-		if len(statistics) != 0 {
-			log.Println("CLI helper - " +
-				"aws cloudwatch get-metric-statistics" +
-				" --metric-name " + metric.Name +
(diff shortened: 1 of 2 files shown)


thomaspeitz:
Thanks for your contribution.

I am a fan of keeping the output json. If somebody needs to log into a file / human readable they can use a service manager for that. Colored output is fancy but not worth that we make the code more complex and adding a library is something i am against as well.

My two cents on that. So i would like to merge some of your additions as e.g. using log.debug is a lot better than my approach.

Can we find a solution between us two and merge that?

Greetings, Thomas

j-nix:
Hi @thomaspeitz.

No problem, I respect your points about adding the extra libraries.
I will amend this PR and just include the log.debug section 👍

thomaspeitz:
Awesome thanks a lot :))

PR #139, opened 2020-03-30
Use get metrics list

daviddetorres:
This PR solves the following problems: 
- When high number of resources, there were many metrics queried, but not all of them had metrics available (for example, in S3, many resources, but maybe only a bunch of them with metrics enabled).
- In ALB there were many queries for target resources that was causing API throttling.  
- The field 'length' of the metrics is being ignored and causes data not being shown

Solution: 
- Rewrite of the scrapeDiscoveryJobUsingMetricData function to reuse list of resources in all the metrics of the same job and only ask for metrics with data available according to GetMetricsList. This prevent throttling when high number of resources and saves petitions of metrics to the API.
- Changed the petition of targeted resources in ALB to reuse the list of metrics. This prevents throttling when high number of ALB resources configured in AWS account.
- Fixed bug in that function that was ignoring the field legnth of the metric and always using 120s
- Replaced the getMetricsData function to use the paginated version
- Changed the number of metrics asked per getMetricsData API request to a configurable flag and set it to 500 by default (current max. metrics for this function)
- Added “ character to be replaced by _ in replaceWithUnderscores function
- Added debugging option for AWS API calls if debug enabled

### DIFF
--- a/abstract.go
+++ b/abstract.go
@@ -121,17 +121,24 @@ func scrapeStaticJob(resource static, clientCloudwatch cloudwatchInterface) (cw 
 	return cw
 }
 
-func scrapeDiscoveryJobUsingMetricData(job job, tagsOnMetrics exportedTagsOnMetrics, clientTag tagsInterface, clientCloudwatch cloudwatchInterface) (awsInfoData []*tagsData, cw []*cloudwatchData) {
+func scrapeDiscoveryJobUsingMetricData(
+	job job,
+	tagsOnMetrics exportedTagsOnMetrics,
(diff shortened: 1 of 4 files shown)


thomaspeitz:
Hey thanks for your big contributions <3 - Should i merge this first or your other PRs? - Can we close other PRs in favour of this one?

Greetings, Thomas

daviddetorres:
Hi Thomas, 

All are important, but this last one solves some important things, the most important one is that it was ignoring the 'length' field in the metric. 

The order in mergin, I leave it up to you :)

PR #144, opened 2020-04-06
Filter only ALB resources

daviddetorres:
Changed resource filter to get only ALB and not all elastic load balancers.


thomaspeitz:
Forgot to say thanks <3 @daviddetorres 

Really appreciate all your work!

daviddetorres:
Always a pleasure!

PR #150, opened 2020-04-20
AppSync auto discovery

eugenetolan:


aws_cloudwatch.go
```diff
@@ -483,6 +485,9 @@ func detectDimensionsByService(service *string, resourceArn *string, fullMetrics
 	switch *service {
 	case "alb":
 		dimensions = queryAvailableDimensions(arnParsed.Resource, getNamespace(service), fullMetricsList)
+		dimensions = queryAvailableDimensions(arnParsed.Resource, getNamespace(service), clientCloudwatch)
```
thomaspeitz, line 488: This looks like an error? - Beside that looks good. Would like to add this :) - Could you remove this line if not needed or explain it. Thank you!

eugenetolan: @thomaspeitz This has been removed.

PR #157, opened 2020-05-31
Add support for Step Function Execution Metrics

smcavallo:
Initial Support for Step Function Execution Metrics
See - https://docs.aws.amazon.com/step-functions/latest/dg/procedure-cw-metrics.html#cloudwatch-step-functions-execution-metrics

AWS/States namespace Dimensions for StateMachineArn 

(note - ActivityArn|LambdaFunctionArn|ServiceIntegrationResourceArn and other non-execution metrics are not supported with this change.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ config.yml
 yet-another-cloudwatch-exporter
 vendor
 dist
+
+.idea/*
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@ config.yml
 yet-another-cloudwatch-exporter
 vendor
 dist
-
-.idea/*
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ YACE is currently in quick iteration mode. Things will probably break in upcomin
   * kafka - Managed Apache Kafka
   * firehose - Managed Streaming Service
   * sns - Simple Notification Service
+  * sfn - Step Functions
 
 ## Image
 
--- a/aws_cloudwatch.go
+++ b/aws_cloudwatch.go
@@ -290,6 +290,8 @@ func getNamespace(service *string) *string {
 		ns = "AWS/Route53Resolver"
 	case "s3":
 		ns = "AWS/S3"
+	case "sfn":
+		ns = "AWS/States"
 	case "sns":
 		ns = "AWS/SNS"
 	case "sqs":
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -2,3 +2,5 @@ config.yml
 yet-another-cloudwatch-exporter
 vendor
 dist
+
+.idea/*
```
thomaspeitz, line 6: Could you please remove the change in the .gitignore files. They are not project specific in my opinion.

You could use https://help.github.com/en/github/using-git/ignoring-files for that :)

smcavallo: @thomaspeitz - removed - makes sense and thank you

smcavallo:
When dimensions are looked up for lambda functions, and this is returned,  
dimensionsWithValue
```
 { Name: "FunctionName", 
Value: "arn:aws:lambda:us-east-1:MY_AWS_ACCOUNT_ID:function:MY_LAMBDA_FUNCTION_1"}]
```
  fullMetricsList
``` {
    Metrics: [{
            Dimensions: [{
                    Name: "FunctionName",
                    Value: "arn:aws:lambda:us-east-1:MY_AWS_ACCOUNT_ID:function:MY_LAMBDA_FUNCTION_1"
                }
            ],
            MetricName: "Invocations",
            Namespace: "AWS/Lambda"
        }...
```		
The existing filtering code ONLY returns metrics both the name and value of the dimensions match the name and value of the full metrics list.  **This works as expected, and is great.**
	
This is what that looks like for "AWS/States":
dimensionsWithValue
```
 { Name: "StateMachineArn", Value: "MY_STATE_MACHINE_NAME"}]
```
 fullMetricsList
```{
    Metrics: [{
            Dimensions: [{
                    Name: "StateMachineArn",
                    Value: "arn:aws:states:us-east-1:MY_AWS_ACCOUNT_ID:stateMachine:MY_STATE_MACHINE_NAME"
                }
            ],
            MetricName: "ExecutionsStarted",
            Namespace: "AWS/States"
        }...
```		
Even though both are named `StateMachineArn` the value is never going to match - one is an arn, the other is a name.  I don't know why "AWS/States" does that, but that is the behavior.
That is why temporarily changing the `StateMachineArn` from the name to the ARN works - it allows the matching in `filterMetricsBasedOnDimensionsWithValues` to succeed.
This is why I took this approach.  It would have a much bigger impact to change the filtering logic in filterMetricsBasedOnDimensionsWithValues to accommodate these "AWS/States" metrics.

It's even odder for other "AWS/States" metrics. 
dimensionsWithValue	
```
 { Name: "StateMachineArn", Value: "MY_STATE_MACHINE_NAME"}]
```
  fullMetricsList
```{
    Metrics: [{
            Dimensions: [{
                    Name: "LambdaFunctionArn",
                    Value: "arn:aws:lambda:us-east-1:MY_AWS_ACCOUNT_ID:function:MY_LAMBDA_FUNCTION_1"
                }
            ],
            MetricName: "LambdaFunctionsStarted",
            Namespace: "AWS/States"
        }...
```		

It would be better to update the filtering logic to account for these in sfn, but as far as I can tell, much bigger changes required for that.

thomaspeitz:
Thanks for your explanations, that helps me a lot!

Will release in the next few days. Sorry for the long delay on this one.

PR #159, opened 2020-06-02
support for multiple regions

goya:
Supports:

- `discovery.job[].regions`
- `static[].regions`

to allow specifying multiple regions.  

This new property overrides `region` only if `region` is not specified.

### DIFF
--- a/abstract.go
+++ b/abstract.go
@@ -16,16 +16,6 @@ var (
 	tagSemaphore        chan struct{}
 )
 
-func getRegions(region string, regions []string) []string {
-	if region != "" {
-		return []string{region}
-	} else if (len(regions)) > 0 {
-		return regions
-	}
-
-	return []string{""}
-}
-
 func scrapeAwsData(config conf) ([]*tagsData, []*cloudwatchData) {
 	mux := &sync.Mutex{}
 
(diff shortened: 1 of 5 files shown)


thomaspeitz:
Let us kill region and make it the new default.

What is your opinion about it?

goya:
added the readme update.

config.go
```diff
@@ -22,6 +22,7 @@ type exportedTagsOnMetrics map[string][]string
 
 type job struct {
 	Region        string   `yaml:"region"`
```
thomaspeitz, line 24: There seems to be still region in there. Can we remove this?

abstract.go
```diff
@@ -16,6 +16,16 @@ var (
 	tagSemaphore        chan struct{}
 )
 
+func getRegions(region string, regions []string) []string {
```
thomaspeitz, line 19: Whole logic is no more needed as far as i can see, if we remove region.

config.go
goya: i'll remove it.  kept it in so it wouldnt be a breaking change, just new functionality - at least for a few versions.

abstract.go
goya: same as above, i'll update it.

thomaspeitz:
Thanks a lot :) Will release the following days!

goya:
this closes issue #160

Issue #164, opened 2020-06-17
Allow customTags for discover jobs


goya:
- allow customTags in discovery jobs not only static jobs

thomaspeitz:
Was merged and released. Thanks again :)

PR #165, opened 2020-06-17
add customTags to discovery jobs

goya:
- add customTags to discover jobs
- ensure valid custom tags with promStringTag

Fix for #164


thomaspeitz:
I like this one. I will merge and release it the following days.

PR #168, opened 2020-06-19
set the correct partition for ASG for gov and cn

smcavallo:
This will construct the new arn using the partition found in the original autoscaling arn instead of defaulting to a partition of `aws`

https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
```
    The following are the supported partitions:
        aws - AWS Regions
        aws-cn - AWS China Regions
        aws-us-gov - AWS GovCloud (US) Regions
```


thomaspeitz:
Thanks a lot for your PR :)

PR #172, opened 2020-06-22
initial ApiGateway support

smcavallo:
Initial support for api gateway.

The challenges with capturing these metrics:

The Arns for api gateway are based off the Ids, NOT the names.
The cloudwatch metric dimensions use the name, not the ID.
This makes it impossible to associate the IDs returned by the resource tagging API with the names metric dimensions without some changes.
See - https://docs.aws.amazon.com/apigateway/latest/developerguide/arn-format-reference.html

What I did to support this:
Describe the api gateways using *WithContext background calls.
The api names are put in a "Matcher" property on the Resource Object.
The thought being that the ID can stay as the ID but a custom property could be used to do the lookup/matching/association between the cloudwatch metric dimension and the tagged object.

There are additional properties in the ARN for stage/method/resource which have unique metrics.
Ex. GET vs POST, qa vs prod - each have their own metrics
See - https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-metrics-and-dimensions.html
See - https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-logging.html
It would be difficult to capture these using the existing additionalDimensions property, because the values are dynamic.

What I did to support this:
These api gateway dimensions are also in the ARNs.
I dynamically add the dimensions by parsing the ARNs.  That allows the matching of dimension counts and dimension names/values to successfully associate and lookup these metrics.
The problem with that is that prometheus won't allow you register a collector where the labels don't exist on all the metrics.
So for all api gateway metrics, empty strings are added to the api metrics for the stage/method/resource if they aren't set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ YACE is currently in quick iteration mode. Things will probably break in upcomin
 * Supported services with auto discovery through tags:
 
   * alb - Application Load Balancer
+  * apigateway - Api Gateway
   * appsync - AppSync
   * cf - Cloud Front
   * dynamodb - NoSQL Online Datenbank Service
--- a/abstract.go
+++ b/abstract.go
@@ -42,6 +42,7 @@ func scrapeAwsData(config conf) ([]*tagsData, []*cloudwatchData) {
 				clientTag := tagsInterface{
 					client:    createTagSession(region, roleArn),
 					asgClient: createASGSession(region, roleArn),
+					apiGatewayClient: createAPIGatewaySession(region, roleArn),
 				}
 				var resources []*tagsData
 				var metrics []*cloudwatchData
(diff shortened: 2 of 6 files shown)


thomaspeitz:
Thanks for this feature! - This was a big part missing.

I will check this out probably on friday. Due to the size i need more time to review!

thomaspeitz:
Could you please fix the merge conflicts? - @smcavallo - Then i merge and test it.

thomaspeitz:
Hey @smcavallo could you please rebase your PR? - Then i would merge it. My plan is to release a new version and i would like to have your PR in there. <3

smcavallo:
There were enough changes were the rebase was rather difficult.
I created a new branch off master and merged my changes into it as part of this PR https://github.com/ivx/yet-another-cloudwatch-exporter/pull/213
I'm closing this PR in favor of https://github.com/ivx/yet-another-cloudwatch-exporter/pull/213

PR #174, opened 2020-06-22
remove duplications between dimensions with and without values

rrusso1982:
A duplication of the settings between `awsDimensions` on a job and the discovered base dimensions for the cloudwatch namespace will cause no metrics to be discovered.  This adds a filter that will remove those duplicates from that list automatically and reduce the possibility for error in the configuration.


thomaspeitz:
Thanks for your contribution.

I need to check this out with some time. Did not get it on first view! :)

Will try to review this on friday.

Issue #166, opened 2020-06-17
Feature Request: SnakeCase and Lowercase label name


goya:
This might get political.  I'm wondering what people's thoughts are on:
- lowercasing all labels names
- snakeCasing all label names

If we want to do this, do we:
1. make it default
2. make it an option, default yes
3. make it an option, default no

goya:
i vote 2!

goya:
i vote 1. it seems to be a convention, and it is used for the metric name.

thomaspeitz:
As aws supports tags e.g. "Environment" and "environment" - I am against it but I keep it open maybe a lot of other users want to see it then we can think of an solution.

daviddetorres:
We should be careful while changing names of the labels. This can affect both dashboards and alerts that users have already configured and in production. 

If any change is done, my opinion is to make it optional, and at least for some versions not to do it by default so users can have some time to adapt their visualizations and alerts.

goya:
i vote 3 now!

thomaspeitz:
Was merged and released :)

PR #175, opened 2020-06-22
add snake case switch

rrusso1982:
To increase "drop in" compatibility with other exporters available currently we need to be able to output labels on the metrics in snake case as opposed to the default camel case.  This PR adds a command line switch for that mode.


goya:
this will fix #166

thomaspeitz:
Could you please add docs @rrusso1982 =)

rrusso1982:
Added documentation

PR #177, opened 2020-06-23
initial unit tests

smcavallo:
would be really useful to have unit tests around core functionality
adding a few simple initial unit tests as a base to start from


thomaspeitz:
Awesome i really love that one!

PR #178, opened 2020-06-23
setting inheritance

rrusso1982:
For commonly repeated metric config elements we can use a setting at the job level that can be overriden by a setting lower down at the metric level.  This adds that ability for:

- period
- addCloudwatchTimestamp


thomaspeitz:
Awesome :) Could you please add docs for that? :)

rrusso1982:
Added some documentation

rrusso1982:
> Could you please fix merge conflicts? @rrusso1982 <3 Thanks!

Should be good to go!

PR #179, opened 2020-06-24
add service name translations

rrusso1982:
When trying to replace an existing exporter it is good to be able to make the replacement a "drop-in" replacement for the old exporter.  This PR adds the ability to translation service names in the metrics output to a list of specified service names instead of their defaults.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,17 @@ YACE is currently in quick iteration mode. Things will probably break in upcomin
 
 ### Top level configuration
 
-| Key       | Description                   |
-| --------- | ----------------------------- |
-| discovery | Auto-discovery configuration  |
-| static    | List of static configurations |
+| Key          | Description                                              |
+| ------------ | -------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


thomaspeitz:
Not merging this one. Thanks for your contribution!

We can set names differently globally for all but let us keep one naming pattern please and not support different schemas.

rrusso1982:
So how would you propose then that we be able to support a migration from one exporter to another?  I am currently trying to adopt this product in my organization but since the metric names it outputs are different than the exporter I am using I cannot just drop this one in place of the old one.  Without some kind of way to make them compatible I am going to need to have hundred of engineers update thousands of dashboards so I am interested to know your thoughts on what we might be able to do?

thomaspeitz:
You can rename during scraping @rrusso1982 - https://medium.com/@alexandrev/how-to-change-the-names-of-your-metrics-in-prometheus-b78497efb5de on prometheus :) - This should solve your issue. Thanks for explaining it.

PR #181, opened 2020-06-27
initial redshift support

smcavallo:
Adds initial support for redshift

See - https://docs.aws.amazon.com/redshift/latest/mgmt/metrics-listing.html


thomaspeitz:
Thanks <3

PR #186, opened 2020-07-01
Remove duplicate type row in metric definition table

calvinbui:
It already appears on L85


thomaspeitz:
Good find :) Thanks for your collaboration! - Really appreciate it!

PR #188, opened 2020-07-07
Fix no info on no values

### LINKED ISSUE #171: Bug: Only shows *_info metrics for metrics with data
Before, all info metrics were shown, and now only the ones with data (as reported in #151). Both are compatible with different scenarios. 

Here there are 2 options: 

1. All *_info metrics are always shown
2. New config parameter to only show info for metrics with data

Can anyone share use cases and thoughts on this?

daviddetorres:
Added optional field in job config 'onlyInfoIfData'
- If true, only adds the info metrics of resources with data.
Fixes #171

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,6 @@ exportedTagsOnMetrics:
 | searchTags           | List of Key/Value pairs to use for tag filtering (all must match), Value can be a regex.   |
 | metrics              | List of metric definitions                                                                 |
 | additionalDimensions | List of dimensions to return beyond the default list per service                           |
-| onlyInfoIfData       | If set to 'true', only adds the info metrics of the resources with data. Optional and 'false' by default.  |
 
 searchTags example:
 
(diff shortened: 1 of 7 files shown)


thomaspeitz:
Great to see you here again <3 

Lets make this default that data is shown and solve it as bug. I would prefer to not make it configurable.

README.md
```diff
@@ -90,6 +90,7 @@ exportedTagsOnMetrics:
 | searchTags           | List of Key/Value pairs to use for tag filtering (all must match), Value can be a regex.   |
 | metrics              | List of metric definitions                                                                 |
 | additionalDimensions | List of dimensions to return beyond the default list per service                           |
+| onlyInfoIfData       | If set to 'true', only adds the info metrics of the resources with data. Optional and 'false' by default.  |
```
thomaspeitz, line 93: Lets make it the default and not make it configurable

daviddetorres: Changed! :+1:

PR #192, opened 2020-07-09
Transit Gateway Attachment Metrics

rhys-evans:
Method to get transit gateway attachment metrics

Does api call so requires the following IAM perms, this is due to the metric requiring both the TransitGateway and TransitGatewayAttachment dimensions to be presented at query time

```json
"ec2:DescribeTags",
"ec2:DescribeInstances",
"ec2:DescribeRegions",
"ec2:DescribeTransitGateway*"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@ config.yml
 yet-another-cloudwatch-exporter
 vendor
 dist
-yace
-.idea/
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -2,3 +2,5 @@ config.yml
 yet-another-cloudwatch-exporter
 vendor
 dist
+yace
+.idea/
```
thomaspeitz, line 6: Could you please add .idea to your global .gitignore. I think this is nothing project specific or am i missing something?

rhys-evans: Removed

PR #212, opened 2020-07-20
set prometheus client to ignore label differences

rrusso1982:
When exporting from cloudwatch you can get into the situation where the same metric key name is exported more than once but with differing sets of label keys attached to the metric.  In the default mode the prometheus client will throw a hard error on this condition.

The clients solution to this situation is to put the client into "unchecked" mode but setting up the Describe function as an empty function.  This internally tot he client disables a number of checks and allows the code creating the metrics to ensure integrity of the metrics.

A goos example of this situation happening is with ECS metrics in which the cpu used for the cluster or for a service have the same metrics name exported but one has an additional label key as part of the exported data.

### DIFF
--- a/prometheus.go
+++ b/prometheus.go
@@ -54,11 +54,7 @@ func NewPrometheusCollector(metrics []*PrometheusMetric) *PrometheusCollector {
 	}
 }
 
-func (p *PrometheusCollector) Describe(descs chan<- *prometheus.Desc) {
-	for _, metric := range p.metrics {
-		descs <- createDesc(metric)
-	}
-}
+func (p *PrometheusCollector) Describe(descs chan<- *prometheus.Desc) {}
 
 func (p *PrometheusCollector) Collect(metrics chan<- prometheus.Metric) {
 	for _, metric := range p.metrics {


rrusso1982:
@thomaspeitz Hey let me know if this one doesn't make sense or if you think maybe we should add an on/off flag for this function.  thanks

thomaspeitz:
Hey @rrusso1982,

won't merge this one. It is from my perspective of course a bug and should be solved but not by disabling a prometheus validation. We need to ensure that we add all labels to all metrics. I am open to another fix of the problem of course.

Sorry to disappoint you here and not merge it.

Greetings, Thomas

rrusso1982:
> Hey @rrusso1982,
> 
> won't merge this one. It is from my perspective of course a bug and should be solved but not by disabling a prometheus validation. We need to ensure that we add all labels to all metrics. I am open to another fix of the problem of course.
> 
> Sorry to disappoint you here and not merge it.
> 
> Greetings, Thomas

I’m willing to do the work to ensure we output unified label set for each metric.  If I did that would it be a worthwhile PR?  I wouldn’t want to waste the development effort.

thomaspeitz:
@rrusso1982 - This is 100% worth a PR!

rrusso1982:
> @rrusso1982 - This is 100% worth a PR!

Ok let me see what I can do!

thomaspeitz:
Awesome thanks!

rrusso1982:
@thomaspeitz I've made a first run through ensuring the consistent labels we dicussed.  The PR is here: https://github.com/ivx/yet-another-cloudwatch-exporter/pull/235.  Enjoy!

PR #235, opened 2020-08-16
label consistency

rrusso1982:
When scraping metrics from cloudwatch there are cases in which metrics for subcomponents of a resource will return with the same metrics name but a different set of labels.  This inconsistency in the metrics labels causes the prometheus client to crash.  This PR attempts to create consistency between the labels of the recorded metrics and the metrics being output.

A good example of this is aws_ecs_cpuutilization_average.  This key is produced for the cluster as a whole and for each service in the cluster.  The cluster level version of the metric does not have the dimension that included the service name.  Hence the cause of the crash.

This adds two major functions:

- recordLabelsForMetric: Creates a record of all labels ever seen for any given metric name
- ensureLabelConsistencyForMetrics: Uses the recorded metrics labels to ensure that every metrics being output has all the possible labels attached to it.

The process to ensure consistency is as follows:

- record labels as they are generated/scraped from cloudwatch.  They are stored in a map by metric key name.
- record the metrics as normal in the metrics buffer.
- after all metrics are recorded re-process the recorded metrics and ensure that they have a consistent set of labels.

We have to record first and ensure consistency later because missing labels can be recorded at any time so we cannot ensure consistency until we have all the data.

### DIFF
--- a/aws_cloudwatch.go
+++ b/aws_cloudwatch.go
@@ -670,6 +670,28 @@ func recordLabelsForMetric(metricName string, promLabels map[string]string) {
     labelMap[metricName] = workingLabelsCopy[:j+1]
 }
 
+func ensureLabelConsistencyForMetrics(metrics []*PrometheusMetric) []*PrometheusMetric {
+    var updatedMetrics []*PrometheusMetric
+
+    for _, prometheusMetric := range metrics {
+        metricName := prometheusMetric.name
+        metricLabels := prometheusMetric.labels
+
+        consistentMetricLabels := make(map[string]string)
+
+        for _, recordedLabel := range labelMap[*metricName] {
+            if value, ok := metricLabels[recordedLabel]; ok {
+                consistentMetricLabels[recordedLabel] = value
+            } else {
+                consistentMetricLabels[recordedLabel] = ""
+            }
+        }
+        prometheusMetric.labels = consistentMetricLabels
+        updatedMetrics = append(updatedMetrics, prometheusMetric)
+    }
+    return updatedMetrics
+}
+
 func migrateCloudwatchToPrometheus(cwd []*cloudwatchData) []*PrometheusMetric {
 	output := make([]*PrometheusMetric, 0)
 
--- a/aws_cloudwatch.go
+++ b/aws_cloudwatch.go
@@ -641,12 +641,34 @@ func getStateMachineNameFromArn(resourceArn string) string {
 	return parsedResource[1]
 }
 
-func promLabelExists(promLabels map[string]string, key string) bool {
-	if _, ok := promLabels[key]; ok {
-		return true
-	} else {
-		return false
+func createPrometheusLabels(cwd *cloudwatchData) map[string]string {
+	labels := make(map[string]string)
+	labels["name"] = *cwd.ID
+	labels["region"] = *cwd.Region
+
+	// Inject the sfn name back as a label
+	switch *cwd.Service {
+	case "sfn":
+		labels["dimension_"+promStringTag("StateMachineArn")] = getStateMachineNameFromArn(*cwd.ID)
+	case "apigateway":
+		// The same dimensions are required on all metrics by prometheus
+		for _, key := range []string{"Stage", "Resource", "Method"} {
+			labels["dimension_"+promStringTag(key)] = ""
+		}
 	}
+
+	for _, dimension := range cwd.Dimensions {
+		labels["dimension_"+promStringTag(*dimension.Name)] = *dimension.Value
+	}
+
+	for _, label := range cwd.CustomTags {
+		labels["custom_tag_"+promStringTag(label.Key)] = label.Value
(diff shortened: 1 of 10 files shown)


jylitalo:
We have a bit same kind of situation in https://github.com/ivx/yet-another-cloudwatch-exporter/blob/c174df726bea3f65aa9319cd15b03703d5216f5f/aws_cloudwatch.go#L651 where we add bunch of labels to all apigateway metrics, so this is not limited to `aws_ecs_*`

aws_cloudwatch.go
```diff
@@ -747,9 +791,12 @@ func migrateCloudwatchToPrometheus(cwd []*cloudwatchData) []*PrometheusMetric {
 				includeTimestamp = false
 			}
 			if exportedDatapoint != nil {
+
+                promLabels := createPrometheusLabels(c)
+                recordLabelsForMetric(name, promLabels)
```
jylitalo, line 796: `go fmt` would probably resolve differences in indentation.

rrusso1982: > We have a bit same kind of situation in
> 
> https://github.com/ivx/yet-another-cloudwatch-exporter/blob/c174df726bea3f65aa9319cd15b03703d5216f5f/aws_cloudwatch.go#L651
> 
> where we add bunch of labels to all apigateway metrics to avoid these problems, so this is not limited to `aws_ecs_*`

ok we should be able to remove that block then since this should handle that case.

rrusso1982:
> We have a bit same kind of situation in
> 
> https://github.com/ivx/yet-another-cloudwatch-exporter/blob/c174df726bea3f65aa9319cd15b03703d5216f5f/aws_cloudwatch.go#L651
> 
> where we add bunch of labels to all apigateway metrics to avoid these problems, so this is not limited to `aws_ecs_*`

ok we should be able to remove that block then since this should handle that case

aws_cloudwatch.go
rrusso1982: much appreciated!

josephreynolds:
Thanks @rrusso1982 this fixed our AWS/Billing queries. This config works with your branch while breaking on master:

```static:
  - namespace: AWS/Billing
    name: TotalEstimated
    regions:
      - us-east-1
    roleArns:
      - "arn:aws:iam::90210:role/assumecw"
    dimensions:
      - name: Currency
        value: USD
    metrics:
      - name: EstimatedCharges
        statistics:
        - Average
        period: 3600
        length: 86400
  - namespace: AWS/Billing
    name: 8675309-per-service
    regions:
      - us-east-1
    roleArns:
      - "arn:aws:iam::90210:role/assumecw"
    dimensions:
      - name: Currency
        value: USD
      - name: ServiceName
        value: "AmazonS3"
      - name: LinkedAccount
        value: 8675309
    metrics:
      - name: EstimatedCharges
        statistics:
        - Average
        period: 3600
        length: 86600```

Master

```bash-5.0$ ./run.sh
{"level":"info","msg":"Parse config..","time":"2020-08-26T08:41:11Z"}
{"level":"info","msg":"Startup completed","time":"2020-08-26T08:41:11Z"}
Name=Currency,Value=USD
Name=Currency,Value=USDName=ServiceName,Value=AmazonS3
Name=Currency,Value=USDName=ServiceName,Value=AmazonS3Name=LinkedAccount,Value=8675309
Name=Currency,Value=USD
panic: descriptors reported by collector have inconsistent label names or help strings for the same fully-qualified name, offender is Desc{fqName: "aws_billing_estimated_charges_average", help: "Help is not implemented yet.", constLabels: {dimension_Currency="USD",dimension_LinkedAccount="8675309",dimension_ServiceName="AmazonS3",name="8675309-per-service",region="us-east-1"}, variableLabels: []}

goroutine 8 [running]:
github.com/prometheus/client_golang/prometheus.(*Registry).MustRegister(0xc00032e050, 0xc00008ff08, 0x1, 0x1)
	/go/pkg/mod/github.com/prometheus/client_golang@v1.7.0/prometheus/registry.go:401 +0xad
main.updateMetrics(0xc00032e050)
	/opt/main.go:88 +0x2ad
main.main.func1(0xc00000e330)
	/opt/main.go:124 +0xbe
created by main.main
	/opt/main.go:121 +0x45b
```

It would be nice to specify whether the user created `name` was a label or part of the metric name. As a proxy exporter is operating blindly, flexible rewrites would be less disruptive.

rrusso1982:
@thomaspeitz Any chance we are going to be able to review this any time soon?  Thanks

thomaspeitz:
Hey @rrusso1982 thanks for your PR :) - Really love it! - Could you rebase please? - Then i merge and try to release this today.

rrusso1982:
> Hey @rrusso1982 thanks for your PR :) - Really love it! - Could you rebase please? - Then i merge and try to release this today.

ok it's rebased and should be all set to go!  thanks much

PR #267, opened 2020-10-21
Fix sorting of datapoints, add test util functions

### LINKED ISSUE #266: Data point sorting not working as intended
The `sortByTimestamp()` method is sorting data points in ascending order (oldest to newest). However, `getDatapoint()` iterates over the slice of data points and returns the first one which matches the statistic it is looking for. This means that the oldest data point is being exported to the output instead of the most recent.

I noticed this because we have a long running Lambda (runs every 30 minutes) that serves as a pass/fail monitor.  If it fails, the Lambda retry mechanic kicks in and it runs again immediately. If the retry passes, I expect the metric to show passing, but it shows failing until the Lambda runs again 30 minutes later.

Botono:
Fixes #266 

Sorting data points in descending order allows `getDatapoint()` to pull the most recent one when it iterates over the slice.

Adds the test helper functions from https://github.com/benbjohnson/testing


thomaspeitz:
Thanks for your PR :)

PR #272, opened 2020-11-03
Add support for exporting metrics from untagged lambdas

### LINKED ISSUE #270: Grabbing metrics from untagged resources?
I've setup yace to export metrics from some lambda functions.  It's working, but doesn't pull in metrics from untagged functions.  Is it possible to force collection of metrics from those functions too?  I'd be happy to deal with AWS IDs in that case.

rvandegrift:
This adds an option to export metrics from untagged resources, and implements it for lambda functions.  This enables using YACE in environments without tagged resources, where you might not control the tags.

Closes #270


thomaspeitz:
Hey @rvandegrift not a fan of this PR initially.

Can you tell me whole reason behind implementing it first?

I would like to understand it. - Tag based finding is one of the main concepts of this exporter which you are breaking with your PR in my opinion.

Greetings, Thomas

rvandegrift:
@thomaspeitz - this feature is mostly intended to support cloud clean-up efforts.

I'm trying to help a team that has years of sprawling lambda development.  We're trying to improve monitoring to ascertain which functions need work and which can be retired, but the environment is very messy.  Most of the functions do not have any tags.  They have different, sometimes broken deployment systems, so adding tags would be a large effort.  And the functions are spread across numerous accounts and regions.

YACE is well-suited since the IAM role-based discovery makes it easy to centralize the data we need.  Numerous accounts & regions makes CloudWatch difficult to use.  Tag-based discovery is better in most other use-cases.

Originally I considered writing a tool to discover & tag lambdas, which would then get picked up by YACE.  I chose this route because:
1. YACE already implements efficient discovery across different roles and regions.  That's the bulk of what I'd need to build.
2. Others probably have some cloud sprawl, and if this would be helpful to them, I'd like to provide the feature.

thomaspeitz:
Hey @rvandegrift 

slept over your answer a night. - I won't support this feature request.

I am very sorry because I see all the work that you did there and totally appreciate that you support the exporter in favour of writing a new tool to tag resources. The reasons behind it that it adds complexity in the code for a rare use case. It should be always the favour to fix the environment and add tags instead of monitoring all untagged resources. Therefore I will close this. You did a great work there and I hope this won't be your last contribution here!

Greetings, Thomas

PR #275, opened 2020-11-07
add RDS DB clusters

goya:
- adds support for RDS cluster metrics.

This is for supporting the new serverless aurora etc which has no instances.

I am amiable to add a new discovery type or just ignore this in discovery and force people to use a static metric


thomaspeitz:
Looks good to me :) Thanks for your PR :)

Issue #276, opened 2020-11-07
Bug in recordLabelsForMetric


goya:
- `recordLabelsForMetric` has a bug where it doesnt copy the previous labels

thomaspeitz:
Thanks for the issue and the fix of it. Appreciate it :))

PR #277, opened 2020-11-07
fix for missing labels

goya:
- fixes a bug around slice copying when dealing with labels

Fix for #276


thomaspeitz:
Thanks a lot :)

PR #280, opened 2020-11-23
adding helm chart for easy kubernetes installation

12ushan:
Simple helm chart which allows you to deploy yet-another-cloudwatch-exporter through helm to your kubernetes cluster.
First contribution so apologies if i've made any mistakes- happy to improve.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ config.yml
 yet-another-cloudwatch-exporter
 vendor
 dist
+.idea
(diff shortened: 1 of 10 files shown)


thomaspeitz:
Hey @12ushan thanks for your contribution 👍 

As you asked for feedback. This looks awesome but i would not add personal stuff like .idea files or .helmignore with a lot of personal special cases to the repository (e.g. .vscode/) beside that it looks good.

I still need to tell you that i am not able to merge this one. I do not want to maintain the helm chart. I would suggest that you add this to your own repository and make a PR just with a link to your repository. This way you can maintain the helm chart and the code keeps separated from deployment. I like this approach more as helm is only one possibility to install it to kubernetes.

Greetings, Thomas

PR #281, opened 2020-11-25
Add support for FIPS compliant endpoints

smcavallo:
See - https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html
This adds support for calling the FIPS compliant endpoints wherever possible.

This PR is an "all or nothing" approach which is largely convention based.  If run with `--fips` all regions will attempt to use the fips compliant endpoints, which may or may not exist for the associated region(s).

The typical pattern is configuration based wherever the region is set, allow a corresponding endpoint to be set.
It seemed too difficult to maintain this in config.file since the service calls behind the region are abstracted.


thomaspeitz:
Hey @smcavallo can you please add only fips endpoints in the PR and not do several things at once. It makes it a lot harder to merge (intending yaml, updating packages, updating go).

smcavallo:
@thomaspeitz - makes sense.  closing this one in favor of https://github.com/ivx/yet-another-cloudwatch-exporter/pull/284
thanks!

PR #284, opened 2020-12-07
use fips flag to enable fips endpoints

smcavallo:
See - https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html
This adds support for calling the FIPS compliant endpoints wherever possible.

This PR is an "all or nothing" approach which is largely convention based. If run with --fips all regions will attempt to use the fips compliant endpoints, which may or may not exist for the associated region(s).

The typical pattern is configuration based wherever the region is set, allow a corresponding endpoint to be set.
It seemed too difficult to maintain this in config.file since the service calls behind the region are abstracted.


thomaspeitz:
Thank you for your new PR :) Appreciate your work and thoughts.

PR #293, opened 2020-12-19
Add support from GameLift

jp:
Add ability to fetch the cloudwatch metrics dynamically.


thomaspeitz:
Thanks for your contribution :)

PR #301, opened 2021-01-11
Auto-Discovery of Dimension Values for Static Jobs

smr-devlin:
## Use Case
My team uses the ELK stack, and we have our Logstash services set up to emit metrics to a custom namespace in cloudwatch.  We would like to use YACE to scrape those metrics.  However, the dimensions of most of the metrics correspond to Kinesis shard IDs.  All in all we have 100ish unique kinesis shards across various streams, with shard IDs liable to change if we reshard any streams.  As it stands we would have to specify all of these shard names in the configuration files of our YACE deployment, and update whenever the streams were resharded.  I think it would be great if YACE could auto-discover the dimension values for us, sparing us the need to create and maintain a huge, unwieldy config file.

I tried to make this PR as a generalizable way to auto-discover dimension values given a namespace and dimension names, and create jobs for those discovered values.  I think other people may be in a similar situation (small number of dimensions, large number of values).    It seems from issues like [190](https://github.com/ivx/yet-another-cloudwatch-exporter/issues/190) and [224](https://github.com/ivx/yet-another-cloudwatch-exporter/issues/224) that auto-discovery for custom namespaces is something people want.  While this doesn't solve the problem of full auto-discovery for custom namespaces, it makes using static jobs for custom namespaces more automated.

## Changes
This introduces the configurable value `populateNamlessDimensions` to static job configs.  If this value is set to `true`, and any `Dimensions` do not have a `Value` specified, those values will be extrapolated.  For each metric specified in the static job, a call will be made to the cloudwatch client's [ListMetrics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_ListMetrics.html) function filtering on the dimensions provided.  Then, for all unique combinations of dimension values returned, a new static job identical to the old in all ways except dimension values will be created for that combination of dimensions.  The new static jobs with dimension values specified will be scraped, and the original static job with unspecified values will not be scraped.
If `populateNamlessDimensions` is set to false no behavior will change.

## Testing
I tested this against my team's AWS account with my own cofig_test.yml file, and with the unit tests provided.  I added unit tests for new functionality.

### DIFF
--- a/abstract.go
+++ b/abstract.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
 	"math"
 	"math/rand"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


zhangela:
@thomaspeitz or @smcavallo (tagging y'all since you've written a lot of the recent commits) - any chance the team might be willing to review this PR? We are really hoping to export per-stream cloudwatch Kinesis metrics to unblock some alerting changes we want to make.

thomaspeitz:
Hey @zhangela - I will review this on Friday.

Would you be so kind and add a config example for your kinesis shards. So i am able to understand it better what you try to resolve. - (https://keybase.io/tsupertramp in cases of private data)

I think I got what you try to solve here but want to be sure. Will look into detail into it but I am unsure if we should add this as functionality or allow a more generic static discovery by providing a discovery url. - So you were able to put almost all your code into a second microservice which then yace just crawls to get the dimensions from. This would allow to keep the code here simpler but still allows you to reach the target. What are you thoughts about such a solution?

Back here on Friday.

Greetings, Thomas

smr-devlin:
@thomaspeitz I added a config example - `custom_static_config_test.yaml`, which I used to test locally.  It scrapes two metrics (`millisBehindLatest` and  `RecordsProcessed`) for each Kinesis shard in a 65 shard stream.  Output looks like the following:
`aws_logstash_s3_millis_behind_latest_average{dimension_Operation="ProcessTask",dimension_ShardId="shardId-000000000335",name="logstash_s3",region="us-east-1"} 0`
`aws_logstash_s3_records_processed_average{dimension_Operation="ProcessTask",dimension_ShardId="shardId-000000000335",name="logstash_s3",region="us-east-1"} 16.673743169398907`

With regards to your proposed solution with a microservice to scrape dimension values, I think that's a viable solution to our problem.  Setting up and running a secondary microservice would be a little messier for us and others with our use case, so I'm hopeful we can do everything in YACE, but if you decided that this code change adds too much complexity to YACE I'd be happy to refactor to that solution.  Thanks for your time and your help with this :)

smr-devlin:
@thomaspeitz any chance you or someone on the team will have time to review this next week?

thomaspeitz:
So I decided to not merge this here. Sorry to disappoint you. I see all your work and thoughts but it is too much complexity in a part which I do not want to maintain. And I want to keep it maintainable here.

Currently thinking how to proceed. The initial approach was to allow autodiscovery through tagging api. So what I would like to suggest is that in manual you can configure a "discoveryEndpoint" which returns json with the values to scrape. This would mean you need a second "microservice" to take care of all. In the long run this would solve issues for more people than you and allows fine granular discovery e.g. just half of the metrics instead of magically all.

I know it is not the best solution for you. I hope you can understand my points here.

If you open source your own endpoint of the autodiscovery, we of course link it in the docs as well :)

Greetings and thanks for your support, Thomas

smr-devlin:
Sounds good, thank you for your time!

PR #317, opened 2021-01-29
Add option to use floating time windows

zqad:
When issue #269 was fixed, the requested "window" (start to end time) was changed to be aligned to 5 minute slots. This turned
out to introduce a regression for when using the metrics as alarms, as a request at e.g. 12:38:31 with length 300s would use a window of 12:35:00 to 12:40:00. This would mean that the actual length (300s) would not be respected, and that any requests does use its own "dynamic delay" to round the time to 5 min intervals. Since metric propagation in AWS is around 2-3 mins, using a 5 min "length" window with 0 delay might return 0 metrics instead of the expected 1-4.

This adds an option to go back to the old behaviours of windows that are not fixed to the 5 min slots.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,9 +63,10 @@ YACE is currently in quick iteration mode. Things will probably break in upcomin
 
 ### Command Line Options
 
-| Option            | Description                                                               |
-| ----------------- | ------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


thomaspeitz:
Thanks a lot :)

PR #331, opened 2021-03-09
Fix resources filtering for alb

nicolasgarfinkiel:
alb metrics are not exposed because the resource gets filtered when constructing target groups resources logic


thomaspeitz:
Could you please rebase? :)

nicolasgarfinkiel:
Hey, latest commits make this change unnecessary as it is now working as expected. When can I expect a new release? Thanks!

PR #345, opened 2021-03-25
Fix on China AWS support 

insectme:
This is from Samsung China server team.  Currently Yace exporter can't work in China, and the logs show "describe resources for region cn-northwest-1: InvalidClientTokenId: The security token included in the request is invalid.\n\tstatus code: 403”.
We found the region need to be specified when session new, if specify after session new, it will not be effective.

The issue is in [https://github.com/ivx/yet-another-cloudwatch-exporter/issues/344](url)


thomaspeitz:
Hey @insectme - Thanks for your PR :) - Could you please rebase? :)

insectme:
some issue when rebase, will recreate the PR. Thanks very much!

Issue #347, opened 2021-03-26
Scrapes running in EKS fail after first scrape


rrusso1982:
I have YACE running in EKS using a service account with an assumed role.  Before the upgrade to 0.26 this worked fine.  After the upgrade it appears that after the first scrape is completed the next scrape returns all NaN data.

Example:
First Scrape:
```aws_dynamodb_consumed_read_capacity_units_average{account_id="XXXXXX",dimension_global_secondary_index_name="",dimension_table_name="entitlements-store-us-production-v1",name="arn:aws:dynamodb:us-east-1:XXXXXXX:table/entitlements-store-us-production-v1",region="us-east-1",tag_environment="production",tag_name="entitlements-store-us-production-v1",tag_role="dynamodb",tag_service="entitlements-store-us"} 0.5 1616779020000```

Second Scrape:
```aws_dynamodb_consumed_read_capacity_units_average{account_id="XXXXXXX",dimension_global_secondary_index_name="",dimension_table_name="entitlements-store-us-production-v1",name="arn:aws:dynamodb:us-east-1:XXXXXX:table/entitlements-store-us-production-v1",region="us-east-1",tag_environment="production",tag_name="entitlements-store-us-production-v1",tag_role="dynamodb",tag_service="entitlements-store-us"} NaN```

**Comment by bryan-kc** [comment]
I have similar issues running on 0.26.3-alpha. 

I have it setup with the configuration found on PromCat (https://promcat.io/apps/aws-alb) 

This is the deployment manifest I am using
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: yace
  namespace: cloudwatch-exporter
spec:
  selector:
    matchLabels:
      app: yace
  replicas: 1
  template:
    metadata:
      labels:
        app: yace
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "5000"
    spec:
      containers:
        - name: yace
          image: quay.io/invisionag/yet-another-cloudwatch-exporter:v0.26.3-alpha
          imagePullPolicy: IfNotPresent
          args:
            - "--config.file=/tmp/config.yml"
          ports:
            - name: app
              containerPort: 5000
          volumeMounts:
            - name: config-volume
              mountPath: /tmp
            - name: yace-credentials
              mountPath: /exporter/.aws/credentials
              subPath: credentials
          securityContext:
            runAsUser: 1000
      volumes:
        - name: config-volume
          configMap:
            name: yace
[...]

PR #350, opened 2021-03-29
Scrapes running in EKS fail after first scrape

rrusso1982:
When the call to get `GetCallerIdentity` is performed outside the metrics scrape functions we can get into the situation where the lookup fails due to token expiration when an inline assume role has been performed.  This most prominently occurs when using a service account assumption from within an EKS cluster.

To alleviate this problem this moves the sts calls within the scraping functions so they are performed in just in time for the accountid to be used as part of the lookups.  

This should be the resolution for #347


Anta-Krumina:
This PR solves our problem when starting up yace. Thanks @rrusso1982 !
Waiting when this is going to be merged in master and ready to use.

thomaspeitz:
Please rebase and I merge and release as fast as possible. Thanks for the fix :)

haarchri:
when we can see the fix in a release ? ;)

rrusso1982:
> Please rebase and I merge and release as fast as possible. Thanks for the fix :)

rebase complete!

haarchri:
we build up a image from master - but it looks like the exporter hangs after first scrape without a message 

```
{"level":"debug","msg":"Metrics scraped.","time":"2021-04-15T14:18:22Z"}
{"level":"debug","msg":"Sleeping at regular sleep interval 172800","time":"2021-04-15T14:18:22Z"}
```

no more message come up since 2 hours 
any ideas ?

PR #354, opened 2021-04-03
autodiscovery missing rds dimensions

sonhador82:
Hello, I've got some issues with thins config file, it's looks like some error in your code with comparing filtered dimensions, so database identifiers missed.
I'll be appreciate, if you accept my fix. thank you.

`discovery:
  exportedTagsOnMetrics:
    rds:
      - Environment
      - Project

  jobs:
  - type: rds
    regions:
      - eu-west-1
    searchTags:
      - Key:  Environment
        Value: production
      - Key: Project
        Value: ge
    period: 30
    length: 300

    metrics:
      - name: CPUUtilization
        statistics: [Average]
`

### DIFF
--- a/aws_cloudwatch.go
+++ b/aws_cloudwatch.go
@@ -315,44 +315,42 @@ func getFilteredMetricDatas(region string, accountId *string, namespace string, 
 			}
 		}
 	}
+
+	var filteredDimensions []*cloudwatch.Dimension
+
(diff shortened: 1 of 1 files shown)


thomaspeitz:
Thanks a lot for your PR - I will check it out in detail. This could take some time as I need to invest more time on this PR to understand the problem.

Beside that, would you please rebase? :)

PR #322, opened 2021-02-22
Make exporter a library.

jeschkies:
Summary:
- Enables users to import the exporter as a library.
- Most methods become stateless.

We would like to use YACE as a library but could not import a `main` package. Thus we propose to move al required code into `pkg`. There is some room for improvement. E.g. operational configuration such as `cloudwatchConcurrency` could be consolidate in one struct.

Depends on #315

PR #359, opened 2021-04-12
Correct build in Dockerfile.

jeschkies:
Summary:
- Follow up to #322.
- Copy all files for build.


haarchri:
thanks - helped to try locally old version is not working ;)

PR #360, opened 2021-04-13
1.Currently Yace exporter can't work in China.

insectme:
2.The error in logs shows "describe resources for region cn-northwest-1: InvalidClientTokenId: The security token included in the request is invalid.\n\tstatus code: 403”
3.Region need to be specified when new session, if do it after session creation there will be error.


insectme:
create a new PR instead of https://github.com/ivx/yet-another-cloudwatch-exporter/pull/345 
rebased

PR #367, opened 2021-04-19
Ensure entries in names map are initialized

friedrichg:
Fixes https://github.com/ivx/yet-another-cloudwatch-exporter/issues/363
and adds 2 test cases.

Please let me know if any adjustment is need.


friedrich-at-adobe:
Ready for Review. Probably a few test cases for ApiGateway https://github.com/ivx/yet-another-cloudwatch-exporter/blob/master/pkg/services.go#L59-L62 would be nice to have but I don't run Apigateway in my accounts, so don't know how the data looks like.

friedrichg:
Ready for review. Probably a test case for Apigateway would be nice to have https://github.com/ivx/yet-another-cloudwatch-exporter/blob/master/pkg/services.go#L59-L62 but I don't run Apigateway in my accounts.

thomaspeitz:
Thanks a lot :)

PR #383, opened 2021-05-07
skip metric if Cloudwatch timestamp is enabled and no metric data

arvidsnet:
if some CloudWatch metric returns no data and addCloudwatchTimestamp is enabled yace returns no timestamp and so, Prometheus is adding current timestamp automatically. This breaks metric ingestion order for metrics which are intermittently unavailable. This simple patch just skips metrics which has no data if addCloudwatchTimestamp is enabled


arvidsnet:
unfortunately this fix was invalid, sorry for that. I created https://github.com/ivx/yet-another-cloudwatch-exporter/pull/421 to fix it.

PR #384, opened 2021-05-13
Use LoadBalancerName dimension for ELB metrics

### LINKED ISSUE #379: ELB labels not being honoured -
0.27.0-alpha

I have found that while my NLB metrics are correctly labeled, my ELB metrics are not:

```yaml
discovery:
  exportedTagsOnMetrics:
    elb:
    - kubernetes.io/service-name
    nlb:
    - kubernetes.io/service-name
  ...
  - type: elb
    regions:
      - some-region
    searchTags:
      - key: kubernetes.io/service-name
        value: .*
```

It's odd because if I look at `aws_elb_info{}`, I can see the correct labels and the searchTags are honoured:

```
aws_elb_info{name="LBARN",tag_kubernetes_io_cluster_clustername="owned",tag_kubernetes_io_service_name="istio-system/istio-ingressgateway"} 0
```

but they are not on the scraped metrics

```
aws_elb_healthy_host_count_minimum{account_id="someaccount",dimension_AvailabilityZone="someaz",dimension_LoadBalancerName="lbname",dimension_Namespace="",dimension_Service="",name="global",region="some-region",tag_kubernetes_io_service_name=""} 2
```

nhinds:
Classic ELBs use `LoadBalancerName` rather than `LoadBalancer` for their dimensions.

Fixes #379

### DIFF
--- a/pkg/services.go
+++ b/pkg/services.go
@@ -290,7 +290,7 @@ var (
 				aws.String("elasticloadbalancing:loadbalancer"),
 			},
 			DimensionRegexps: []*string{
-				aws.String(":loadbalancer/(?P<LoadBalancer>.+)$"),
+				aws.String(":loadbalancer/(?P<LoadBalancerName>.+)$"),
 			},
 		}, {
 			Namespace: "AWS/ElasticMapReduce",


uvaisibrahim:
When we do this the metric name is being changed from `aws_elb_latency_average` to `aws_elb_az_latency_average`. (v0.23.0-alpha). Why this is happening?

nhinds:
Version `v0.23.0-alpha` does not have the change in this PR, it was released in `v0.28.0-alpha`.

Using the current latest version `v0.33.0-alpha` and the following `config.yml`, I get `aws_elb_latency_average` metrics:
```
apiVersion: v1alpha1
discovery:
  jobs:
  - type: elb
    regions:
      - ap-southeast-2
    metrics:
      - name: Latency
        statistics:
        - Average
```

It does not appear that the metric names are changed, at least not for me.

If you're having problems, I would recommend raising a new issue rather than commenting on a year-old PR. In your new issue, you should include your full configuration, what versions you are comparing, what you expect to see, and the actual metrics that are being returned so that the problem can be diagnosed.

PR #392, opened 2021-06-01
Fix nilToZero ignoring issue for static job

khanhhongpham:
Hi,
I noticed the nilToZero option doesn't work for static jobs, then I tried to fix it by this change. I already see it works as expected on my environment. Could the team review this change?
Thank you


thomaspeitz:
Hey, should be fixed in the newest release (probably released today)

PR #407, opened 2021-06-25
Fix logic in dimensions for Transit Gateway Attachments

rhys-evans:
Requires both TransitGateway and TransitGatewayAttachement Dimensions


thomaspeitz:
Thanks a lot

PR #410, opened 2021-07-06
Label metrics based on TargetGroup

kyleplant:
For some metrics under the one namespace, there may be multiple resources involved. Using Application/ELB as an example, we have LoadBalancers and TargetGroups. I would like to label my metrics using the Tags from the TargetGroup specified in the dimension, otherwise use LoadBalancer tags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ config.yml
 yet-another-cloudwatch-exporter
 vendor
 dist
+.vscode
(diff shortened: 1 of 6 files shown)


thomaspeitz:
Hey, need to close this one. Sorry! Really appreciate your contribution but for me it is too much complexity. Please use different tag keys (so the overwrite wont happen).

Wont merge this one.

dlahn:
This one would really help us.

dlahn:
@thomaspeitz Your solution doesn't seem to work. If I have a tag on my LB, say `tag1=true`, and then `tag2=true` on my target group, if I set `tag2` as the exported tag, then it just comes through as an empty string in the final metric, as it just adds to the top LB as empty.

PR #438, opened 2021-09-02
Fix race condition setting end time for discovery jobs.

cristiangreco:
This fixes a data race happening when goroutines calling GetMetricData
attempt to update the `endtime` variable.


thomaspeitz:
Thanks a lot!

Issue #346, opened 2021-03-26
Feature request: Add live reload support for configuration files


Diggs27:
Prometheus and other open source monitoring tools have a feature that allow you to hot reload configuration files. 
In the prometheus project, making a `POST` request to the  '/-/reload` endpoint will attempt to validate the config file and reload changes. This feature request is for the yace project to support a similar feature.

antoniomerlin:
In this feature what will be the status of metrics scraping. Should it scrape on endpoint hit.

cristiangreco:
This has been implemented.

PR #444, opened 2021-09-16
Feature: reload endpoint for hot load changes from config file

antoniomerlin:
In this PR an endpoint is added for loading hot changes without restarting the container for #346 . Just making POST request to `/-/reload` latest config changes are available in metrics.

Let me know if any changes are required in this.

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -110,7 +110,21 @@ func main() {
 		</body>
 		</html>`))
 	})
-	http.HandleFunc("/-/reload", s.reload(ctx))
+	http.HandleFunc("/-/reload", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		log.Println("Parse config..")
+		if err := config.Load(configFile); err != nil {
+			log.Fatal("Couldn't read ", *configFile, ": ", err)
+		}
+		if *decoupledScraping {
+			s.decoupled(ctx)
+		} else {
+			s.scrape(ctx)
+		}
+	})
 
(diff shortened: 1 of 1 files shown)


cmd/yace/main.go
```diff
@@ -170,3 +171,22 @@ func (s *scraper) scrape(ctx context.Context) (err error) {
 	log.Debug("Metrics scraped.")
 	return nil
 }
+
+func (s *scraper) reload(ctx context.Context) func(http.ResponseWriter, *http.Request){
```
thomaspeitz, line 175: Lets put this function into the pkg/config.go. Would like to keep main.go small.

Can you please verify if the code was run in gofmt? https://pkg.go.dev/cmd/gofmt

thomaspeitz (changes requested):
Thanks for your PR :) lets make some small changes and get this merged.

thomaspeitz:
@antoniomerlin If you get this fixed till Thuersday morning, I get this into the next release.

Else I schedule it for next week :)

Greetings :)

cmd/yace/main.go
antoniomerlin: Hi thanks for reviewing, yes the code is run in gofmt

antoniomerlin:
Hi i removed that reload as separate function and added directly inside http handlefunc, as for change requested to move this function inside config.go currently the way i am doing this is loading the config file again and calling the scrape function for making the config changes available in next metric scrape.
```
log.Println("Parse config..")
if err := config.Load(configFile); err != nil {
	log.Fatal("Couldn't read ", *configFile, ": ", err)
}
if *decoupledScraping {
	s.decoupled(ctx)
} else {
	s.scrape(ctx)
}
```
What are your thought about this approach

PR #451, opened 2021-09-25
Refactor tagsData (now taggedResource) and add test cases.

cristiangreco:
This is a minor refactoring around the code in `aws_tags.go`:
- rename struct to taggedResource
- move methods next to the receiver
- cleanup taggedResource fields (`ID`->`ARN`, no pointers)
- add tests

No user-facing functionality is changed here.


thomaspeitz:
Thanks a lot will release with next version :)

PR #452, opened 2021-09-27
Lint code with GolangCI tool.

cristiangreco:
This add a GolangCI setup and runs it in GitHub actions.

Only a few linters are enabled, and in this PR we also fix some minor
issues spotted by the tool.

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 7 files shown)


cmd/yace/main.go
```diff
@@ -167,10 +166,10 @@ func (s *scraper) decoupled(ctx context.Context, cache exporter.SessionCache) {
 	}
 }
 
-func (s *scraper) scrape(ctx context.Context, cache exporter.SessionCache) (err error) {
+func (s *scraper) scrape(ctx context.Context, cache exporter.SessionCache) {
```
cristiangreco, line 169: The linter was complaining about unchecked return value. For this method as it is now there's not much we can do in case of error, so I'm just removing it from signature.

thomaspeitz:
Thank you :)

PR #458, opened 2021-10-01
Return error on getFullMetricsList instead of log.Fatal

matthewnolf:
This PR updates the `getFullMetricsList` func to return an error - in the case the `ListMetricsPages` CloudWatch call fails. 

This error is then logged by the caller, in `getMetricDataForQueries` with additional context.

The current behaviour `log.Fatal`s in this scenario, causing the running processes to exit with code **1**. Errors from the CloudWatch API should be expected, and handled gracefully.


thomaspeitz:
Lovely. Thanks!

PR #477, opened 2021-11-09
Allow setting default statistics in job config

surminus:
We can already set default values for most values, but we can't for statistics, and force the user to explicitly set them. I cannot
immediately see why we can't also set statistics with a default value.

This copies the logic for other values, but instead of setting a default, it fails if neither a default or an explicit value is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,10 @@ discovery:
       - eu-west-1
     length: 900
     delay: 120
+    statistics:
+      - Minimum
+      - Maximum
+      - Sum
     searchTags:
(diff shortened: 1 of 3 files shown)


thomaspeitz:
Awesome! - Thanks a lot.

thomaspeitz:
@surminus - Hey could you PR the docs as well. - Sad to have it as feature without users knowing it. - Thanks a lot <3

surminus:
Done :) https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/483

PR #480, opened 2021-11-14
Fix the calculation of start and end times for GetMetricData

### LINKED ISSUE #479: GetMetricData time calculations are broken for some use cases
Hi all!

First of all, thank you for building this exporter with the detached functionality!  This is just what I need so multiple teams can scrape the exporter without incurring additional costs for CW requests.

I've been testing this exporter to use for ALB and NLB metrics.  As others have pointed out, I found that metrics did not match what's shown on the AWS CloudWatch console. After doing some debugging, I found that, depending on the settings for _period_, _length_, and _delay_, the exporter calculates the start/end times of the GetMetricData requests incorrectly. The current method of calculating the start/end time may work for some very specific use cases, but for others it's quite broken.

To figure out what's going on, I added log entries to make the start/end time of the GMD requests visible for debugging. Using the settings below, I found every interval the start/end time would go 2 minutes **backwards** in time. Behavior like this could be the cause of other issues that have been submitted stating that retrieved metrics are incorrect.

Settings tested which result in GMD requests going backwards in time:
```
delay=120
length=120
period=120
scraping-interval=120
```

Below, you can see the logs where start time goes backwards on each iteration. In addition to GMD requests going backwards in time, it was also retrieving metrics that were already over 5 minutes old, even though my delay is set to 2 minutes. This is due to the hard coded rounding to the nearest 5 minutes [here](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/add2e47efdb34c450f46135443b659b5f2992810/pkg/aws_cloudwatch.go#L127).
```
{"level":"info","msg":"createGetMetricDataInput: start time: 2021-11-09T15:41:00-06:00","time":"2021-11-09T15:46:34-06:00"}
{"level":"info","msg":"createGetMetricDataInput: end time: 2021-11-09T15:43:00-06:00","time":"2021-11-09T15:46:34-06:00"}
{"level":"info","msg":"createGetMetricDataInput: start time: 2021-11-09T15:39:00-06:00","time":"2021-11-09T15:48:43-06:00"}
{"level":"info","msg":"createGetMetricDataInput: end time: 2021-11-09T15:41:00-06:00","time":"2021-11-09T15:48:43-06:00"}
{"level":"info","msg":"createGetMetricDataInput: start time: 2021-11-09T15:37:00-06:00","time":"2021-11-09T15:50:43-06:00"}
{"level":"info","msg":"createGetMetricDataInput: end time: 2021-11-09T15:39:00-06:00","time":"2021-11-09T15:50:43-06:00"}
```

I believe the core of the issue may be that future CW requests are building the start/end time based on the end time of the previous request.  This may work for some uses cases, but depending on the period/delay/length settings, it won't work for all uses cases.  I believe the solution is to recalculate start/end time using wall clock time for every iteration.  The hard coded 5 minute rounding of time results in configurations with period/length/delay times shorter than 5 minutes to unnecessarily pull metrics older than the use may expect as well (as in my case).

csquire:
Fixes #479

This PR tries to address the problems I outlined in #479.

Here's what I did:
- Updated to always use the wall clock time to calculate start/end of GetMetricData requests
- Remove the hard coded 5 minute rounding of time currently done prior to finding start/end times and instead round based on the shortest configured period for the scrape job
- Add a config option to the job which allows users to easily control the rounding of the time. This is especially useful when someone has a very long period such as 1 day (for S3 metrics) but doesn't want to the time rounded to the nearest day.
- Remove the `floating-time-window` flag. The same functionality can be achieved by setting the new `roundingPeriod` configuration option to 0.
- The `fips` arg was unused by some functions, so I removed it from all function params where it wasn't used (just basic cleanup work, I can put it back if you need it?)
- Extract constants for default period, delay and length settings

Looking forward to your feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,6 @@ We will contact you as soon as possible.
 | Option               | Description                                                                       |
 | -------------------- | --------------------------------------------------------------------------------- |
 | labels-snake-case    | Causes labels on metrics to be output in snake case instead of camel case         |
-| floating-time-window | Use a floating start/end time window instead of rounding times to 5 min intervals |
 
 ### Top level configuration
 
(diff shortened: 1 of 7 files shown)


thomaspeitz:
Awesome will review this Thuersday Morning! - Thanks a lot.

I am so full of client work currently / switched to being freelancer that I did not find to fix anything yet. So happy to see this PR and work on it.

thomaspeitz:
Hey would you be so kind and fix the linting issues @csquire ? :)

csquire:
@thomaspeitz I think it should be good now for the linting, IntelliJ failed me. 😉 
I also rebased the branch.

PR #482, opened 2021-11-17
Issues form templates

NickLarsenNZ:
I was searching through issues but labels aren't used much. They're also painful to add by hand, so I thought it might be a good idea to use Issues Forms:

- Prompt users for a type of issue
- Auto label by issue type

It is also easier on the issue requester because it can prompt them for the necessary information.

### Issue types

Issue Types

#### Bug

Bug Form

#### Feature request

Feature Request Form


thomaspeitz:
<3 Thats such a lovely contribution. Thanks a lot!

PR #490, opened 2021-12-03
add liveness and readiness probes

aleslash:
Proposing standard liveness and readiness probes to bether work with helm.
Used the same structure as in prometheus and other exporters

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -136,12 +136,12 @@ func main() {
 
 	http.HandleFunc("/-/healthy", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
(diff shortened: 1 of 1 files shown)


thomaspeitz:
Lets use `/healthz` seeing it most in the docs of k8s e.g. https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ and as long as we are not doing anything special in those functions:

Keep it with only /healthz - We can improve that in the future if we are doing real checks e.g. if tagging api is reachable or something like that.

Thanks a lot for the PR and the raised issue here!

thomaspeitz (changes requested):
see comment

aleslash:
@thomaspeitz made it a plain /healthz as the k8s docs.

PR #498, opened 2021-12-13
added some more info in the contribute guideline

aleslash:
Added some info for helping the new contributors


thomaspeitz:
Thanks always loving docs updates.

Sorry got me right before vacation. Now back!

PR #512, opened 2022-01-26
main: cancel `ctx` on config reload

### LINKED ISSUE #499: [BUG] Config reload leaks goroutines

### Current Behavior

When the `/reload` endpoint is hit, a new goroutine for scraping is kicked off. The goroutine running the old scrape from the old configuration does not stop, resulting in multiple scrapes and scrapes running on undesired config.

### Expected Behavior

When the `/reload` endpoint is hit, the old goroutine is signalled to stop.

### Steps To Reproduce

1. With any config, run with `--scrape-interval 10`
2. In a separate terminal, run `for i in `seq 1 10`; do curl -X POST localhost:5000/reload; sleep 1; done`
3. Look at the logs in the first terminal to see that there are many goroutines running scrapes.	

### Anything else?

Happy to work on a patch to upstream if fixing is desired.

gumpt:
This cancels the context passed to the background scrape job when the
endpoint to reload configuration is hit. This prevents a goroutine leak
and scrapes based on an old config running.

Fixes #499.

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -118,8 +118,8 @@ func main() {
 	s := NewScraper()
 	cache := exporter.NewSessionCache(config, fips)
 
-	ctx := context.Background() // ideally this should be carried to the aws calls
-
+	// TODO: Pipe ctx through to the AWS calls.
+	ctx, cancelRunningScrape := context.WithCancel(context.Background())
 	go s.decoupled(ctx, cache)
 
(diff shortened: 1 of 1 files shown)


gumpt:
@thomaspeitz unsure how to request a review, and would appreciate one! 😅

thomaspeitz:
@gumpt Thanks a lot for your PR :)

Currently full in client projects and not having much free "open source" time.
Hope to get this released fast.

PR #530, opened 2022-02-21
fix targetgroup arn lookup #529

### LINKED ISSUE #529: [BUG] alb, nlb targetgroup metrics not joinable on name with aws_(alb,nlb)_info metric

### Current Behavior

The metrics emitted by yace for alb and nlb regarding targetgroup metrics e.g. aws_alb_healthy_host_count_minimum are not possible to correlate with the respective aws_(alb,nlb)_info metric. (https://github.com/nerdswords/yet-another-cloudwatch-exporter#query-examples-without-exportedtagsonmetrics)

The problem is that the name label (arn) for the info metric is using the targetgroups ARN while the targetgroup related metric e.g. aws_alb_healthy_host_count_minimum is using the loadbalancers ARN.

aws_alb_healthy_host_count_minimum{name="arn:aws:elasticloadbalancing:eu-west-1:1111:loadbalancer/app/alb-external/asdfasdf"}

aws_alb_info{name="arn:aws:elasticloadbalancing:eu-west-1:1111:targetgroup/asdf-asdf/xasdf", tag_Port="9999"}

This does not work to join with:
(aws_alb_healthy_host_count_minimum + on (name) group_left(tag_Port) aws_alb_info)

### Expected Behavior

Being able to join the targetgroup related metrics.

### Steps To Reproduce

Use targetgroup related metrics like:
```
"discovery":
  "jobs":
  - "enableMetricData": true
    "metrics":
    - "length": 300
      "name": "HealthyHostCount"
      "period": 300
      "statistics":
      - "Minimum"
    - "length": 300
      "name": "UnHealthyHostCount"
      "period": 300
      "statistics":
      - "Maximum"
    "regions":
    - "eu-central-1"
    "type": "alb"
```
Try to join targetgroup related metrics.

### Anything else?

_No response_

domcyrus:
This provides a way to retain the metric name arn and tags if at least one dimension filter matches. The current implementation is always using the last dimension lookup whatever that might be. This change does retain the metrics if it already matched previously.

With this change you can fix #529 using the following config:
```
"discovery":
  "jobs":
  - "enableMetricData": true
    "metrics":
    - "length": 3600
      "name": "ProcessedBytes"
      "period": 300
      "statistics":
      - "Sum"
    "regions":
    - "eu-west-1"
    "type": "alb"
  - "enableMetricData": true
    "metrics":
    - "length": 300
      "name": "HealthyHostCount"
      "period": 300
      "statistics":
      - "Minimum"
    - "length": 300
      "name": "UnHealthyHostCount"
      "period": 300
      "statistics":
      - "Maximum"
    searchTags:
      - key: Port
        value: .*
    "regions":
    - "eu-west-1"
    "type": "alb"
```

Note using the alb type twice. We use it with the tag we expect to be present on the targetgroup ARN.

### DIFF
--- a/pkg/aws_cloudwatch.go
+++ b/pkg/aws_cloudwatch.go
@@ -302,6 +302,7 @@ func getFilteredMetricDatas(region string, accountId *string, namespace string, 
 	}
 	for _, cwMetric := range metricsList {
 		skip := false
+		alreadyFound := false
 		r := &taggedResource{
 			ARN:       "global",
 			Namespace: namespace,
@@ -309,9 +310,12 @@ func getFilteredMetricDatas(region string, accountId *string, namespace string, 
 		for _, dimension := range cwMetric.Dimensions {
 			if dimensionFilterValues, ok := dimensionsFilter[*dimension.Name]; ok {
 				if d, ok := dimensionFilterValues[*dimension.Value]; !ok {
-					skip = true
+					if !alreadyFound {
+						skip = true
+					}
 					break
 				} else {
+					alreadyFound = true
 					r = d
 				}
 			}


thomaspeitz:
Thanks :)

PR #535, opened 2022-02-25
allow passing in dimension label prefix

yashsarma:
**Premise** we are currently migrating away [cloudwatch exporter](https://github.com/prometheus/cloudwatch_exporter) to yace. We have our current dashboards and alarms in our codebase that expect labels to not have the "dimension_" default prefix that is current introduced by Yace. 

**Solution** add a top level config variable to allow optionally overriding the default ("dimension_") metric label prefix.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,8 @@ jobs:
       if: github.event.action != 'published'
       run: exit 78
     - uses: actions/checkout@master
+    - name: Build docker image
+      run: docker build -t yace --build-arg VERSION=${{ github.event.release.tag_name }} .
     - name: Log into docker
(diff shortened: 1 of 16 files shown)


yashsarma:
cc @thomaspeitz, would you mind taking a look please?

thomaspeitz:
Hey @yashsarma - Thanks for the PR and sorry for the late reply.

I won't merge this one here. - It is easier to migrate but I would prefer less complexity / less configurability in the exporter.

We can publish a migration guide to help people migrating but I would propose to keep the "dimension" prefix.

codedumper1:
useful feature, can't understand why it is rejected, especially when full code PR is provided.

PR #560, opened 2022-04-13
Register AWS API counter metrics in main

kgeckhart:
This PR adjusts how the AWS API Counter metrics are exposed and registered. When using YACE as a library it can be useful to register them separately from the registry provided to `UpdateMetrics`.

This PR also adds a section to the README about using YACE as a library and adds some godocs to the library entry point.


thomaspeitz:
Awesome thanks :)

PR #564, opened 2022-04-22
Accept a logger interface in update.go

kgeckhart:
This PR introduces a logging interface to be used in the process of updating metrics. This is the last global dependency which exists in the library entrypoint. This will make it a lot easier for library consumers to plugin their own logging implementation which can carry extra contextual information for logs.

I did also translate all the `Printf` use cases to `Error` because all of them felt like error cases vs info.

I plan to do a follow up on this to switch log messages which embed data in the message to [structured logs](https://github.com/sirupsen/logrus#fields).


thomaspeitz:
Looks interesting, will review as soon as possible :) - Can you please tag me as soon as you are ready with it?

kgeckhart:
@thomaspeitz this should be good to go now. After I found the issue with the logger not being passed I wanted to spend a little more time testing.

thomaspeitz:
Awesome thanks a lot!

PR #568, opened 2022-04-27
Allow VPC Endpoint (Service) resources to be discovered without tag api

### LINKED ISSUE #567: [BUG] For VPCEndpoint(Service) resources, YACE tag requirements clash with CloudFormation limitations.

### Current Behavior

YACE requires resources to be tagged for discoverability, but some resource types, like VPC Endpoints and VPC Endpoint Services, cannot be tagged through CloudFormation (https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/196). In organizations that primarily use CloudFormation to create resources, this clash between CFn and YACE functionality can make YACE difficult to use.

The workarounds are to manually tag the resources (which creates drift between the live resources and declarative definitions), or to use a Lambda-based CFn custom resource to tag the resources (which is tedious and adds a lot of complexity to the CFn stack).

### Expected Behavior

In these scenarios where there is a reasonable explanation for why resources may not be tagged, I think it makes sense to use other APIs for resource discovery. YACE already supports discovering resources without the tag API.

### Steps To Reproduce

_No response_

### Anything else?

_No response_

jutley:
Fixes #567

This PR updates discovery of these resources to not use the AWS tag API. While the tag API works and these resources can be tagged, they cannot be tagged through CloudFormation, so this PR aligns with CloudFormation's limitations.

I can imagine that the maintainers of this project may want to only deviate from using the tags API when strictly necessary. If that is the case then this PR can be closed, but I think this change is worth considering.

### DIFF
--- a/pkg/services.go
+++ b/pkg/services.go
@@ -558,6 +558,32 @@ var (
 			DimensionRegexps: []*string{
 				aws.String(":vpc-endpoint/(?P<VPC_Endpoint_Id>.+)"),
 			},
+			ResourceFunc: func(ctx context.Context, iface tagsInterface, job *Job, region string) (resources []*taggedResource, err error) {
+				pageNum := 0
+				return resources, iface.ec2Client.DescribeVpcEndpointsPagesWithContext(ctx, &ec2.DescribeVpcEndpointsInput{},
+					func(page *ec2.DescribeVpcEndpointsOutput, more bool) bool {
+						pageNum++
+						ec2APICounter.Inc()
+
+						for _, vpce := range page.VpcEndpoints {
+							resource := taggedResource{
+								ARN:       fmt.Sprintf("arn:aws:ec2:%s:%s:vpc-endpoint/%s", region, *vpce.OwnerId, *vpce.VpcEndpointId),
+								Namespace: job.Type,
+								Region:    region,
+							}
+
+							for _, t := range vpce.Tags {
+								resource.Tags = append(resource.Tags, Tag{Key: *t.Key, Value: *t.Value})
+							}
+
+							if resource.filterThroughTags(job.SearchTags) {
+								resources = append(resources, &resource)
+							}
+						}
+						return pageNum < 100
+					},
+				)
+			},
 		}, {
 			Namespace: "AWS/PrivateLinkServices",
 			Alias:     "vpc-endpoint-service",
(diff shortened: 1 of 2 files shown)


pkg/services.go
```diff
@@ -552,20 +552,72 @@ var (
...
+				aws.String(":vpc-endpoint-service/(?P<Service_Id>.+)"),
+			},
+			// CloudFormation doesn't support tagging VPCEndpointService resources, so use a ResourceFunc instead of a ResourceFilter
+			ResourceFunc: func(ctx context.Context, iface tagsInterface, job *Job, region string) (resources []*taggedResource, err error) {
+				firstRequest := true
+				var nextToken *string
+				for firstRequest || nextToken != nil {
+					ec2APICounter.Inc()
+					firstRequest = false
+					response, err := iface.ec2Client.DescribeVpcEndpointServicesWithContext(ctx, &ec2.DescribeVpcEndpointServicesInput{MaxResults: aws.Int64(1000), NextToken: nextToken})
```
jutley, line 598: There isn't any function that fetches all pages of this API like there is with the other APIs, so I had to handle the pagination on my own.

cristiangreco:
> This PR updates discovery of these resources to not use the AWS tag API. While the tag API works and these resources can be tagged, they cannot be tagged through CloudFormation, so this PR aligns with CloudFormation's limitations.

I don't think I'm on the same page here, why would we want to be aligned with this limitation? Resources might have been created/tagged using tools other than CloudFormation.

jutley:
@cristiangreco I will first note that at this point, we have started using hacky workaround using Lambda to tag these resources. Unideal, but we no longer need the functionality provided by this PR.

For clarity, this change isn't introducing a new limitation to YACE, but rather introduces a workaround to deal with CloudFormation's shortcomings. If the resources are created and tagged using a different tool, these changes do not interfere with that. But if the resources are created in an environment where tagging isn't possible (such as CloudFormation which many people use), then the resources can still be used in YACE without using a CFn custom resource to add tags.

thomaspeitz:
To this point the tagging API is something we want to keep as a core element of the exporter. We do not intend to change this behaviour at this point.

Super sorry for the shortcomings of Cloudformation and that it makes your ops life harder :(

If you every build something to retag resources (for yace usage) feel free to PR a link to your repository. You are not the only one struggling with tags beeing a core element of the xporter.

PR #606, opened 2022-06-27
Return early if no tagged resources are found.

cristiangreco:
This adds a check in discoveryJobs after clientTag.get() so that if no
tagged resource is present we return early instead of continuing with
ListMetrics request.


thomaspeitz:
Awesome thanks!

PR #621, opened 2022-07-14
Fix searchTags and bad dimension name

femiagbabiaka:
This commit resolves https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/365.

In cases in which the name of the dimensions for a given metric didn't
match a dimension in our searchTags list, we would pass a default global query
for all metrics under a given namespace into our list, returning metrics
we meant to filter. A test case was added to cover.

In addition, I decorated a debug log entry that was useful in finding
this bug.

### DIFF
--- a/pkg/aws_cloudwatch.go
+++ b/pkg/aws_cloudwatch.go
@@ -307,19 +307,24 @@ func getFilteredMetricDatas(region string, accountId *string, namespace string, 
 		if len(dimensionNameList) > 0 && !metricDimensionsMatchNames(cwMetric, dimensionNameList) {
 			continue
 		}
-		for _, dimension := range cwMetric.Dimensions {
-			if dimensionFilterValues, ok := dimensionsFilter[*dimension.Name]; ok {
-				if d, ok := dimensionFilterValues[*dimension.Value]; !ok {
+
+		for _, dimension := range cwMetric.Dimensions { // loop over the returned dimensions for our current metric
+			if dimensionFilterValues, ok := dimensionsFilter[*dimension.Name]; ok { // if the name of the metric matches the name of a dimension we're filtering against, continue
+				if d, ok := dimensionFilterValues[*dimension.Value]; !ok { // if the value of our metric doesn't match the value of the dimension we're looking for and we haven't already found our metric in the previous iteration of the loop, skip
 					if !alreadyFound {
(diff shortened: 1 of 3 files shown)


pkg/aws_cloudwatch.go
```diff
@@ -307,19 +307,24 @@ func getFilteredMetricDatas(region string, accountId *string, namespace string,
 		if len(dimensionNameList) > 0 && !metricDimensionsMatchNames(cwMetric, dimensionNameList) {
 			continue
 		}
-		for _, dimension := range cwMetric.Dimensions {
-			if dimensionFilterValues, ok := dimensionsFilter[*dimension.Name]; ok {
-				if d, ok := dimensionFilterValues[*dimension.Value]; !ok {
+
+		for _, dimension := range cwMetric.Dimensions { // loop over the returned dimensions for our current metric
```
cristiangreco, line 311: Comments are a bit overwhelming here, maybe you could write just 1-2 lines _before_ the loop to explain what the code is intended to achieve at high level?

pkg/aws_cloudwatch_test.go
```diff
@@ -61,6 +61,120 @@ func TestSortyByTimeStamp(t *testing.T) {
 	require.Equal(t, expectedDataPoints, sortedDataPoints)
 }
 
+func Test_getFilteredMetricDatasBadData(t *testing.T) {
```
cristiangreco, line 64: Instead of having a separate test, could you merge this example with the test table in `Test_getFilteredMetricDatas`?

cristiangreco (commented):
Thanks for your contribution @femiagbabiaka! I left a couple of comments for some small improvements.

pkg/aws_cloudwatch_test.go
femiagbabiaka: done!

pkg/aws_cloudwatch.go
femiagbabiaka: done!

femiagbabiaka:
hello again @cristiangreco! no rush on this, but I wanted to get a sense of when you'd be free to review and ship a release! we're waiting on the fix internally so I wanted to get a gauge on if I should ship the patch on a forked copy until there's more time.

Issue #224, opened 2020-07-31
Scraping metrics from custom namespaces


alex-tsbk:
Hi.
I'm wondering if there is a way to take advantage of auto-discovery but still use custom namespaces for some of the metrics.
It's not really clear to me if it is already supported, so I'm raising this question here.

The most common use-case is obtaining MemoryUtilization metrics, which is not the default EC2 metrics dimension.

I would like to have structure similar to this one when defining my metrics:

```
discovery:
  exportedTagsOnMetrics:
    ec2:
  jobs:
  - type: ec2
    regions:
      - us-east-1
    searchTags:
      - Key: Application
        Value: .*
    awsDimensions:
      - InstanceId
    addCloudwatchTimestamp: true
    metrics:
      - name: CPUUtilization
        period: 60
        length: 120
        statistics:
          - Average
      - name: MemoryUtilization
        namespace: MyCustomNamespace
        period: 60
        length: 120
        statistics:
          - Average
```

So the expected behavior, in this case, would be not using AWS/EC2 namespace, but looking up metrics in my custom namespace: `MyCustomNamespace`

Underlying aws cli commands looks like this:
```
aws cloudwatch get-metric-statistics --metric-name MemoryUtilization --start-time 2020-07-30T20:22:00Z --end-time 2020-07-30T20:27:00Z --period 300 --namespace MyCustomNamespace --statistics Average --region us-east-1 --dimensions Name=InstanceId,Value=i-0123456789012345
```

I would put a PR but I have 0 proficiency in Go language. 

P.S. Amazing work & amazing plugin!

pluvio-phile:
Hi,
I wanted to leave a comment about this feature request as well. I recently tried this exporter and it is amazing how fast and efficient it is.  Please consider prioritizing this feature request if it is possible since application related metrics go to custom namespaces in cloudwatch.

Thank you so much for the outstanding work!

thomaspeitz:
Thanks a lot for your warm words folks,

I would like to avoid to put custom namespace logic into autodiscovery. As this logic is already to complex and I need to refactor a lot.

What I could think of (shorterm) would be the following.
Not quite sure if it solves your problem as if your dimensions do only contain the instanceId and not the tag Name it would not help much as it would be hard to monitor it.

```
customMetrics:
  metrics:
    # Scrape all metrics in namespace
    - namespace:
        - MyService
    # Scrape one metric in a namespace
    - namespace:
        - MyService
      metrics:
[...]

PR #623, opened 2022-07-20
add suport to custom namespaces with their dimensions

arielly-parussulo:
## What?
- add a new property called **customMetrics** that allows users to set custom namespaces.
- the logic will get the list of metrics that the user added in the configuration file, get their datapoints and dimensions and export them to Prometheus.
- The configuration is based on the one mentioned in [224 issue](https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/224). 
- The configuration file would look like the below:

```yaml
apiVersion: v1alpha1
sts-endpoint: eu-west-1
customMetrics:
  - name: customMetrics
    namespace: CustomEC2Metrics
    regions:
      - us-east-1
    metrics:
      - name: cpu_usage_idle
        statistics:
          - Average
        period: 300
        length: 300
        nilToZero: true
      - name: disk_free
        statistics:
          - Average
        period: 300
        length: 300
        nilToZero: true
```

## Why?
- in my company we had a scenario where we needed to collect the metrics from a custom namespace. 

## How?
- By adding a new YAML property in the configuration file to describe the custom namespace metrics.
- By collecting the description of the metrics for that custom namespace and their respective data points from CloudWatch.

**OBS.: I didn't add the unit tests yet as I don't know if you will suggest changes in the code. When it looks ok I can add them.**


thomaspeitz:
Looks awesome! - If you add some tests to this, we can merge and release it this week.

suryam27:
@thomaspeitz my environment has a bunch of custom metrics and we are transitioning from prometheus cloudwatch exporter to YACE, any update on when we can start using this feature :)

arielly-parussulo:
> 
@thomaspeitz, I added some unit tests based on what the project currently have. Please let me know if we need anything else to merge this PR. :)

suryam27:
@thomaspeitz just checking if this is planned for merge anytime soon.

datsabk:
@cristiangreco Could we get a look at this PR too. It would close the loop on Amazon Managed Prometheus metrics by enabling us to fetch metrics for AWS/Usage namespace.

pkg/config.go
```diff
@@ -14,10 +14,11 @@ const defaultLengthSeconds = int64(300)
...
 type ScrapeConf struct {
-	ApiVersion string    `yaml:"apiVersion"`
-	StsRegion  string    `yaml:"sts-region"`
-	Discovery  Discovery `yaml:"discovery"`
-	Static     []*Static `yaml:"static"`
+	ApiVersion    string           `yaml:"apiVersion"`
+	StsRegion     string           `yaml:"sts-region"`
+	Discovery     Discovery        `yaml:"discovery"`
+	Static        []*Static        `yaml:"static"`
+	CustomMetrics []*CustomMetrics `yaml:"customMetrics"`
```
cristiangreco, line 21: What do you think of `customNamespace` as a name instead?

pkg/config.go
```diff
@@ -108,9 +129,16 @@ func (c *ScrapeConf) Load(file *string) error {
 	return nil
 }
 
+func (c *ScrapeConf) IsConfigFileEmpty() bool {
```
cristiangreco, line 132: Not worth having a separate function here, maybe just add `&& c.CustomMetrics == nil` to the original check.

pauruiz:
Sorry to ask, but what is the advantage of using this new custom metrics section, instead of creating an entry on the static section of the config? We do that extensively with custom namespaces, and works fine.

I can clearly see that this job may be reused by the scrapper to make a discovery of custom metrics in the future, but today it seems like it creates a redundant way of defining exactly the same.

datsabk:
@pauruiz - `Auto-discovery` is the keyword here. Static configuration doesn't support auto-discovery of resources. For instance, for AWS/Usage metrics, we require to search for Managed Prometheus workspaces based on the "Required Dimensions" criteria. This can only be done if auto-discovery is enabled. Else, we'd need to manually add dimension name and value for each workspace which is not productive.

pauruiz:
@datsabk - Thanks a lot for the quick response - We have sorted the auto-discovery on the static config by scheduling a cron job that would launch a new re-deploy when a new metric is detected, and using the delay property to capture the although lost event from this newly detected metric (although as a drawback this recaptured event will appear with a significant delay on the grafana system).

I have tried 0.42.0-alpha on my system, and it does not seem to auto-discover the metrics. As per your words, current implementation seems to capture only the dimensions for the given metrics (I have to agree with you, the auto-discovery of the dimensions is an improvement from the original state), but that does not looks like the optimal solution.

On the #224 proposal, an empty custom metric like this one:

`
customMetrics:
  metrics:
    # Scrape all metrics in namespace
    - namespace:
        - MyService
`

Would capture all metrics inside the custom namespace (like the discovery section does), but the implemented solutions seems to rely on the list of metrics, thus only half of the work is done?
Is there a plan to implement that?
May we be doing the config wrong? Or misunderstanding the proposal?

The config we tried was as simple as:
`
customMetrics:
  - name: fdc385296a106746ae2ed133ddf91ed3
    namespace: lambda-metrics
    regions:
    - eu-west-1
    statistics:
    - Average
    - Maximum
    - Sum
    period: 60
    length: 300
    nilToZero: true
`

Am I naive to expect this configuration enough to capture Average, Maximum and Sum dimensions for all metrics inside the custom namespace lambda-metrics?

PR #663, opened 2022-09-15
Add dimensions regex for AmazonMQ.

### LINKED ISSUE #643: Searchtags do not work for Amazonmq

### Current Behavior

The following configmap created for amazonmq collection
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: prometheus-yace-exporter
  namespace: openshift-monitoring
data:
  config.yml: |-
    ---
    apiVersion: v1alpha1
    sts-endpoint: <region>
    discovery:
      exportedTagsOnMetrics:
        mq:
          - Name
          - cluster
      jobs:
        - type: AWS/AmazonMQ
          regions:
            - <region>
          searchTags:
            - key: cluster
              value: <system>
          metrics:
            - name: RabbitMQDiskFree
              statistics:
                - Average
              period: 60
              length: 300
            - name: RabbitMQDiskFreeLimit
              statistics:
                - Average
              period: 60
              length: 300
```

however, all mq brokers are sent to prometheus:

amazonmq

on the logs it can be seen that it skips the non-matching resource:

{"account":"xxxxxxxxx","arn":"","job_type":"AWS/AmazonMQ","level":"debug","msg":"Skipping resource because search tags do not match","region":"eu-north-1","time":"2022-08-20T03:09:54Z"}

however, at some point this ignored and it ends up sending all brokers to prometheus

This does not happen for other resources like RDS, elasticsearch, etc

### Expected Behavior

Amazonmq should respect the searchtags parameter and only send to prometheus the matching brokers/nodes

### Steps To Reproduce

create a configmap that uses searchtags for amazonmq

### Anything else?

_No response_

cristiangreco:
This should fix an issue with AmazonMQ when using `searchTags` which makes unwanted metrics being included in the scrape even though the corresponding resource was filtered out due to `searchTags` constraints.

Due to the current implementation, ideally all services should have regexs defined. I'll followup with further PRs for those services that need be updated.

Fixes #643

### DIFF
--- a/pkg/services.go
+++ b/pkg/services.go
@@ -126,6 +126,9 @@ var (
 			ResourceFilters: []*string{
 				aws.String("mq"),
 			},
+			DimensionRegexps: []*string{
+				aws.String("broker:(?P<Broker>[^:]+)"),
+			},
 		}, {
 			Namespace: "AWS/AppSync",
 			Alias:     "appsync",


femiagbabiaka:
@cristiangreco Can you explain a bit more? I saw that the change I made got reverted for breaking things with more than one dimension. Is the solution to the original problem to define a regex for every service dimension?

PR #692, opened 2022-10-19
add aws ses support

spartan2015:


README.md
```diff
@@ -408,6 +408,41 @@ static:
         - Minimum
         period: 60
         length: 300
+  - namespace: AWS/SES
```
cristiangreco, line 411: Rather than this full example, maybe just add SES to the list of supported services at the top of the file.

cristiangreco: Ignore this comment, the service is already there. So this is just a fix to add the `ResourceFilters`?

marekhanzlik: Does the `ResourceFilter` affect this? 
https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/738

cristiangreco:
Closing this PR as the change shouldn't be needed anymore since #738 has been fixed.

PR #695, opened 2022-10-24
Import CloudWatch mixin.

jeschkies:


mixin/dashboards/ebs.libsonnet
```diff
@@ -0,0 +1,253 @@
...
+    multi=true,
+    sort=common.sortAlphabeticalAsc,
+  )
+)
+.addPanels(
+  [
+    grafana.panel.text.new(
+      title='Info',
+      content=|||
+        Showing metrics only for AWS resources that have tags assigned to them. For more information, see [Amazon CloudWatch Metrics for Amazon EBS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using_cloudwatch_ebs.html).
```
cristiangreco, line 61: Hmm this copy only makes sense for `discoveryJobs` right? I'm wondering how we could make these dashboards more generic yet customisable. Maybe we should find a way to make better use of `config.libsonnet`.

jeschkies: True but I think this is a good first start.

cristiangreco (approved):

PR #703, opened 2022-10-27
Support extraEnv in Helm chart

gnadaban:
This PR adds support for defining extra environment variables.

A use-case where this is necessary is when using `eks.amazonaws.com/role-arn` service account annotation to define the "default" identity the Deployment, which is used to assume various other IAM roles to support multiple accounts. 

In that case, the IAM policies and trust policy must be configured such that the "default" role the deployment is running as can assume other roles. For this, the AWS session name must be deterministic. AWS Session names can be defined via the `AWS_ROLE_SESSION_NAME` env var. 

Example trust policy:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "AllowAssumingRole",
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:sts::<AccountID>:assumed-role/<Role Name>/<Session Name>"
            },
            "Action": "sts:AssumeRole"
        }
    ]
}
```

The corresponding [AWS reference](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html) `Assumed-role session principals` section explains it as:

> When you specify an assumed-role session in a Principal element, you cannot use a wildcard "*" to mean all sessions. Principals must always name a specific session.

### DIFF
--- a/charts/yet-another-cloudwatch-exporter/templates/deployment.yaml
+++ b/charts/yet-another-cloudwatch-exporter/templates/deployment.yaml
@@ -45,9 +45,13 @@ spec:
           volumeMounts:
             - name: vol-yet-another-cloudwatch-exporter
               mountPath: /config
+        {{- if or (.Values.aws.secret.name) (and .Values.aws.aws_secret_access_key .Values.aws.aws_access_key_id) (and (.Values.extraEnv)) }}
+          env:
+        {{- if .Values.extraEnv }}
+          {{- toYaml .Values.extraEnv | nindent 12 }}
+        {{- end }}
         {{- if not .Values.aws.role }}
         {{- if .Values.aws.secret.name }}
-          env:
             - name: AWS_ACCESS_KEY_ID
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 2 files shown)


charts/yet-another-cloudwatch-exporter/templates/deployment.yaml
```diff
@@ -45,9 +45,13 @@ spec:
           volumeMounts:
             - name: vol-yet-another-cloudwatch-exporter
               mountPath: /config
+        {{- if or (.Values.aws.secret.name) (and .Values.aws.aws_secret_access_key .Values.aws.aws_access_key_id) (and (.Values.extraEnv)) }}
```
cristiangreco, line 48: This might be a bit hard to read... what if we change to always have an `env:` block, even if it ends up being empty? e.g. something like

```
env:
  {{ if set of vars }}
    something
  {{ end }}
  {{ if another set of vars }}
    something else
  {{ end }}
  {{ if extraEnv }}
    extra envs
  {{ end }}
```

harshajith: I also wanted to add proxy settings within the env variables to connect to AWS apis through cooperate environment. So better if we can get this reviewed ASAP. Thanks.

abhisar007: Hi, i think it would actually good to introduce an env variable to add proxy for an Enterprise environment. Do we have any fix to introduce proxy in Helm cahrt? or any workaround ?

cristiangreco:
Closing this PR as extra envs are already supported in the helm chart now.

PR #712, opened 2022-11-01
Helm chart update for optional test-connection pod

nikosmeds:
Add Helm chart support for disabling the `test-connection` pod.

We don't use Helm directly, but rather consume the charts with Tanka. We'd like the option to remove this test pod. Please let me know if there's a preferred method for doing so - thanks!


cristiangreco (approved):
LGTM

PR #714, opened 2022-11-02
Helm chart update to set priorityClassName

nikosmeds:
Allow Helm users to set a non-default `priorityClassName` for the YACE deployment.


cristiangreco (approved):
LGTM 👍

PR #746, opened 2022-12-23
Fix not to refer to loop variable in a goroutine

masshash:
This PR fixes to refer to the value passed to the argument instead of referencing the loop variable in the goroutine that scrapes custom metrics.
This fix prevents unpredictable behavior such as getting duplicate metrics.


cristiangreco (approved):
Thanks!

PR #752, opened 2023-01-05
Refactor code into separate packages.

cristiangreco:
Move existing files into a few separate packages. This should improve testability and make future refactorings easier.


pkg/services/aws_tags.go
```diff
@@ -1,4 +1,4 @@
-package exporter
+package services
```
kgeckhart, line 1: This file feels a little out of place here since it's not helping define supported services like services.go. `TaggedResource` and `FilterThroughTags` is used by services but I I'm not sure the services pkg owns the definition of a `TaggedResource`. I think it might fit better with the `job` pkg, WDYT?

pkg/config/config.go
```diff
@@ -92,12 +85,20 @@ type Dimension struct {
...
-func (c *ScrapeConf) Load(file *string) error {
+func (r *Role) ValidateRole(roleIdx int, parent string) error {
+	if r.RoleArn == "" && r.ExternalID != "" {
+		return fmt.Errorf("Role [%d] in %v: RoleArn should not be empty", roleIdx, parent)
+	}
+
+	return nil
+}
+
+func (c *ScrapeConf) Load(file *string, validSvc func(string) bool) error {
```
kgeckhart, line 101: This refactor stood out, it feels a bit odd to allow a consumer to provide a func which validates the services. I think I would expect a defined way to do that as the default behavior of loading the config. What are your thoughts here?

cristiangreco: Totally agree. I've resorted to this approach to avoid a dependency loop: almost all other pkgs depend on `config`, including `services`, so couldn't make `config` depend on `services`.

Might move `services.go` into `config` pkg... a bit extreme but might make sense on a level. What do you suggest?

kgeckhart: Ah yeah I didn't think about the depenency loop 😢 . I think moving `services.go` to the `config` pkg sounds like a good plan for now.

cristiangreco: Second thought, related to the comment on `aws_tags.go` above. Moving `services.go` to `pkg/config` make the latter depend on `pkg/services` (or whatever other package where `aws_tags.go` is placed). Again a loop.

A possible alternative is a code refactoring where `services.go` is split in something like: all the const stuff (names, aliases, etc) is moved to a struct of some sort in `pkg/config`, while the `{Filter,Resource}Func` and everything else that depends on `aws_tags.go` definitions is kept in `services.go`. (plus some additional changes here and there to deal with the fact `SupportedServices` is being split)

pkg/exporter/update.go
```diff
@@ -0,0 +1,66 @@
+package exporter
```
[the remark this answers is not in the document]
cristiangreco: This is fixed now.

pkg/config/config.go
kgeckhart: 👍 makes sense to let this clear and adjust further after

PR #754, opened 2023-01-09
Fix to use length of metrics level on customNamespace.

### LINKED ISSUE #729: [BUG] StartTime equals EndTime for GetMetricData on customNamespace

### Current Behavior

Looks like there is an issue with GetMetricData on customNamespace.
No matter what parameters of lenght , delay, period, roundingPeriod I try, start_time & end_time are allways set to the same value.

Most simple config: 

```
customNamespace:
  - name: xxxxx
    namespace: "xxxxx"
    regions:
      - eu-west-1
    period: 60
    metrics:
      - name: xxxxxxx
        statistics:
          - Average
        length: 600
```

Results in: 

"start_time":"2022-11-21T14:57:00+00:00",
"end_time":"2022-11-21T14:57:00+00:00",

-----------------------------------------------------
DEBUG: Response monitoring/GetMetricData Details:
---[ RESPONSE ]--------------------------------------
...
    <Code>ValidationError</Code>
    <Message>The parameter StartTime must not equal parameter EndTime.
</Messag>

### Expected Behavior

lenght is used as expected on customNamespace 

### Steps To Reproduce

see current behavior 

### Anything else?

customNamespace  is not processed at all with latest release https://github.com/nerdswords/yet-another-cloudwatch-exporter/releases/tag/v0.42.0-alpha , although its already documented. 

Described behavior can be reproduced with latest code in master branch (commit 7c7562cd064f708eed26131ed87bffbeedd83792)

masshash:
This PR fixes #729
customNamespaces always used length on job level and not respect length on metrics level.
With this change, use specified metrics level length.

Fixed with the following two commits:
- Fix to use length of metrics level on customNamespace
- Set default values for metric parameters (fix [this](https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/729#issuecomment-1374349002))

And other:
- Log a error message when metrics are not specified on customNamespace

If #752 is merged, I'm going to rebase this branch.

### DIFF
--- a/pkg/job/abstract.go
+++ b/pkg/job/abstract.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/config"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/logger"
-	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/model"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/services"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/session"
 )
@@ -199,13 +198,8 @@ func scrapeStaticJob(ctx context.Context, resource *config.Static, region string
 	return cw
 }
 
-func getMetricDataInputLength(job *config.Job) int64 {
-	length := model.DefaultLengthSeconds
-
-	if job.Length > 0 {
-		length = job.Length
-	}
-	for _, metric := range job.Metrics {
+func getMetricDataInputLength(metrics []*config.Metric) (length int64) {
+	for _, metric := range metrics {
 		if metric.Length > length {
 			length = metric.Length
 		}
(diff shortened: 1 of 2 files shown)


cristiangreco:
@masshash thanks for the PR, could you please rebase your branch?

pkg/job/abstract.go
```diff
@@ -199,13 +198,8 @@ func scrapeStaticJob(ctx context.Context, resource *config.Static, region string
...
 }
 
-func getMetricDataInputLength(job *config.Job) int64 {
-	length := model.DefaultLengthSeconds
-
-	if job.Length > 0 {
-		length = job.Length
-	}
-	for _, metric := range job.Metrics {
+func getMetricDataInputLength(metrics []*config.Metric) (length int64) {
```
cristiangreco, line 201: Could you please use an explicit `length :=0` instead of the named return param? 🙏

masshash:
Updated (b24716848a8e2efedac2278f9a616b90c85fda44), thanks for the review @cristiangreco

PR #759, opened 2023-01-16
Refactor list of supported services and filter funcs.

cristiangreco:
Extract out supported services list into the config package, and leave filtering funcs in the services package.

Followup of #752


pkg/services/filters.go
```diff
@@ -0,0 +1,260 @@
...
+type serviceFilter struct {
+	// ResourceFunc can be used to fetch additional resources
+	ResourceFunc func(context.Context, TagsInterface, *config.Job, string) ([]*TaggedResource, error)
+
+	// FilterFunc can be used to the input resources or to drop based on some condition
+	FilterFunc func(context.Context, TagsInterface, []*TaggedResource) ([]*TaggedResource, error)
+}
+
+// filterFuncs maps a service namespace to (optional) serviceFilter
+var filterFuncs map[string]serviceFilter = map[string]serviceFilter{
```
kgeckhart, line 31: Nit: wdyt about naming this serviceFilters?

pkg/services/filters.go
```diff
@@ -0,0 +1,260 @@
...
+	"AWS/ApiGateway": {
```
kgeckhart, line 32: WDYT about adding a test to filters_test.go which checks that all keys produce a non-nil service when calling `config.SupportedServices.GetService(`? Might help prevent some awkward bugs later.

kgeckhart (approved):
LGTM just a small suggestion

pkg/services/filters.go
thepalbi: +1, to keep aligned both maps

thepalbi (approved):
LGTM

pkg/services/filters.go
cristiangreco: Just to doble-check my understanding, do you mean checking that every key used here is a vaild service name as defined by `config.SupportedServices`?

kgeckhart: That correct, ideally it would catch if a someone was trying to apply serviceFilters to an incorrect name.

PR #760, opened 2023-01-16
feat: bump helm chart

rasta-rocket:
Hi @thomaspeitz @cristiangreco 👋 

Thanks for having enabled the github pages 🙏 

Now that the helm repo is officially available, let's bump and use the latest version of the exporter 🤩 

I take also the opportunity to :
- add a generated README, it's not perfect but its a start
- trigger the helm release pipeline when the helm chart is modified (usually when a contributor bump the chart version in Chart.yaml)

Don't hesitate to ping me if you need more details 👍 

Cheers 😉

### DIFF
--- a/charts/yet-another-cloudwatch-exporter/Chart.yaml
+++ b/charts/yet-another-cloudwatch-exporter/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: yet-another-cloudwatch-exporter
 description: Yet Another Cloudwatch Exporter
 type: application
-version: 0.8.0
-appVersion: "v0.42.0-alpha"
+version: 0.9.0
+appVersion: "v0.43.0-alpha"
 home: https://github.com/nerdswords/yet-another-cloudwatch-exporter
 sources:
 - https://github.com/nerdswords/yet-another-cloudwatch-exporter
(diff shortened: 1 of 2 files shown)


cristiangreco (approved):
Thanks!

.github/workflows/release_helm.yml
```diff
@@ -2,8 +2,10 @@ name: Build, test and publish Helm chart
 
 on:
   push:
-    tags:
-    - 'v*'
+    branches:
+    - 'master'
+    paths:
+    - "charts/yet-another-cloudwatch-exporter/Chart.yaml"
```
cristiangreco, line 8: On a second thought, might be better to leave this as it was before.

Merging this PR would cause a new helm release for yace `v0.43.0-alpha`, but we've merged a few PRs already since `v0.43.0-alpha`.

When compiling the changelog for the next release, I'll make sure that the chart is updated as well (we can write a quick CI check for that as well).

rasta-rocket: Ok 😄 

Quick question: on which kind of modification would like to release the helm chart ?

rasta-rocket: Ok 👍 
It's done in #763 
Last question regarding that new PR do you want me to keep the current behavior where the pipeline is triggered on git tag ?

cristiangreco (approved):

PR #794, opened 2023-02-04
Add support for extra env variables

harshajith:
Adds support for adding extra env variables. This is required under different conditions, i.e adding proxy settings etc

If there are no values defined, it will keep empty env: block


cristiangreco:
@harshajith the helm chart has been moved out of this repo, could you please open your PR against https://github.com/nerdswords/helm-charts?

cristiangreco:
Fixed in https://github.com/nerdswords/helm-charts/pull/13

PR #815, opened 2023-02-16
Add support for Container, queue, and database metrics for MWAA

millin:
https://docs.aws.amazon.com/mwaa/latest/userguide/accessing-metrics-cw-container-queue-db.html


LeeS-NW:
Definitely in support of this PR - this new namespace will be really useful to add given it is distinctly different from the existing AmazonMWAA namespace.

PR #831, opened 2023-02-28
Refactor CW metrics to resource association logic and add tests

thepalbi:
This PR is the first of two that will fix https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/821. This first PR refactors the logic to contain the actual fix, and adds the failing test scenarios so we can track when the issue is fixed.

### DIFF
--- a/pkg/job/associator.go
+++ b/pkg/job/associator.go
@@ -1,16 +1,24 @@
 package job
 
 import (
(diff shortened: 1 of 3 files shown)


cristiangreco (approved):
Great! 🚀

PR #834, opened 2023-03-02
Return aws_*_info metrics for jobs with no cloudwatch metrics

### LINKED ISSUE #639: [BUG] Missing NLB metrics after upgrade to 0.38-alpha

### Current Behavior

Upgraded yet-another-cloudwatch-exporter from 0.32 to 0.38 and found NLB metrics that were previously being scraped missing. No changes to the NLB resources itself and no indication of errors in YACE logs.

Attached prometheus screengrab shows no NLB metrics being returned after upgrading YACE on 8th Augus
Screenshot 2022-08-15 at 11 10 07
t

### Expected Behavior

NLB metrics being scraped as before. 

### Steps To Reproduce

YACE version 0.38.0-alpha running in kubernetes cluster.
Config for NLB which has stayed the same throughout the upgrade:
      - type: nlb
        regions:
          - eu-west-1
        metrics:
          - name: UnHealthyHostCount
            statistics:
              - Maximum
            period: 300
            length: 600

### Anything else?

_No response_

nhinds:
The `aws_*_info` metrics can be useful even when cloudwatch has no data, e.g. using `aws_nlb_info` for detecting NLBs even with AWS is not reporting any `UnHealthyHostCount` metrics

Fixes #639

### DIFF
--- a/pkg/job/scrape.go
+++ b/pkg/job/scrape.go
@@ -48,7 +48,7 @@ func ScrapeAwsData(
 					jobLogger = jobLogger.With("account", *result.Account)
 
 					resources, metrics := runDiscoveryJob(ctx, jobLogger, cache, metricsPerQuery, tagSemaphore, discoveryJob, region, role, result.Account, cfg.Discovery.ExportedTagsOnMetrics)
-					if len(resources) != 0 && len(metrics) != 0 {
+					if len(resources) != 0 || len(metrics) != 0 {
 						mux.Lock()
 						awsInfoData = append(awsInfoData, resources...)
 						cwData = append(cwData, metrics...)


nhinds:
The failing lint check seems to be a timeout. I don't think I have permissions to retry the check:
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.51.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
  
  Error: golangci-lint exit with code 4
  Ran golangci-lint in 71649ms
```

cristiangreco:
@nhinds I believe this is fixed in `master` now, would you be able to test it please?

nhinds:
This does not seem to be fixed in `master`. The code in pkg/job/scrape.go still [only adds `aws_*_info` metrics if at least one metric was retrieved from cloudwatch as part of the discovery job for the role/region](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/master/pkg/job/scrape.go#L50).

In case it helps, the ["No metrics data found" message is logged](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/6d9841f725428f304c6872284a061c3b87a2ff8c/pkg/job/discovery.go#L86), not ["No tagged resources made it through filtering"](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/6d9841f725428f304c6872284a061c3b87a2ff8c/pkg/job/discovery.go#L75) or ["Couldn't describe resources"](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/6d9841f725428f304c6872284a061c3b87a2ff8c/pkg/job/discovery.go#L77). This means that at least one item is in the `resources` slice, but it gets ignored by `pkg/job/scrape.go` because no metrics for those resources were found.

Adding dummy metrics that you know are always present (e.g. `ProcessedBytes` for NLBs) is still required as a workaround to force some metrics to be returned from the region.

Removing the `if len(metrics) != 0` condition makes the `aws_*_info` metrics always show up if their `searchTags` matched. I could update this PR to remove the `len(metrics)` condition completely if that's okay, or I could rebase this PR on top of `master` to change it to `len(resources) != 0 || len(metrics) != 0`?

It's not super clear to me why the condition is present, other than maybe trying to avoid locking the mutex if a discovery job found no data - but it doesn't seem like we would have a high amount of contention around that mutex from empty scrapes during normal operation.

cristiangreco:
@nhinds there's been a very similar contribution on this topic: https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1130

I then remembered about this old PR of yours. Sorry for leaving it go stale.

The PR above introduces a feature flag for the new logic. Would you be able to test it out and check it solves your use case?

nhinds:
I tried switching to SDK v2 at the same time and ran into #1029 where all of the metrics for my current account (with no assumed role) disappeared.

When sticking with the SDK v1, the `always-return-info-metrics` flag works how I would expect though. Thank you for following up.

PR #856, opened 2023-03-14
Allow discovery jobs to return result even if there are no resources

### LINKED ISSUE #257: Configuration to get AWS billing metrics using YACE
Hi, I am trying to get billing metrics using the YACE exporter. Does it support billing metrics? If yes, can someone please help me with the configuration to get them?

### LINKED ISSUE #738: [BUG] AWS SES resource cant have TAG assigned, so is not scraped anymore

### Current Behavior

AWS SES metrics are not scraped anymore, because they do not contain any TAG, but you can't assign tags to SES

### Expected Behavior

Should scrape SES anyway, even without tags

### Steps To Reproduce

_No response_

### Anything else?

`v3.36.1-alpha` was last that version was working  
i suppose it was caused by 
```
Return early if no tagged resources are found. by @cristiangreco in https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/606
```
that was merged in version `v3.36.2-alpha`

kgeckhart:
There are a few services such as `AWS/Billing`, and `AWS/SES` which are not resources based. This excludes them from being able to be run via a discovery job. This PR attempts to fix that via removing some early exits in the event resources are not found. 

Resource discovery and metrics discovery are only tied together via a loose ability to associate data between the two. If a metric can't be linked back to a resource it's tagged as "global" already, https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/1159de1154138efffc224447c552a54618391f3c/pkg/job/discovery.go#L190-L199. This to me indicates, as long as we didn't expect to find resources it's okay that none were found.

Resolves #738
Resolves #257

### DIFF
--- a/pkg/apitagging/client.go
+++ b/pkg/apitagging/client.go
@@ -57,8 +57,10 @@ func NewClient(
 func (c Client) GetResources(ctx context.Context, job *config.Job, region string) ([]*model.TaggedResource, error) {
 	svc := config.SupportedServices.GetService(job.Type)
 	var resources []*model.TaggedResource
+	shouldHaveDiscoveredResources := false
 
 	if len(svc.ResourceFilters) > 0 {
+		shouldHaveDiscoveredResources = true
 		inputparams := &resourcegroupstaggingapi.GetResourcesInput{
 			ResourceTypeFilters: svc.ResourceFilters,
 			ResourcesPerPage:    aws.Int64(100), // max allowed value according to API docs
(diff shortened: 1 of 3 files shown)


pkg/apitagging/client.go
```diff
@@ -119,5 +122,11 @@ func (c Client) GetResources(ctx context.Context, job *config.Job, region string
 		}
 	}
 
+	if shouldHaveDiscoveredResources && len(resources) == 0 {
+		return nil, ErrExpectedToFindResources
+	}
```
kgeckhart, line 127: This does produce a behavioral change, previously no resources was an info and it will now be an error. I'm not sure this is a bad thing though since it's likely user error that no resources were found when expected.

pkg/apitagging/client.go
```diff
@@ -119,5 +122,11 @@ func (c Client) GetResources(ctx context.Context, job *config.Job, region string
...
+
 	return resources, nil
 }
+
+var ErrExpectedToFindResources = errors.New("expected to discover resources but none were found")
```
thepalbi, line 132: nit: can we move the error definition towards the top of the file?

cristiangreco (approved):

PR #862, opened 2023-03-27
Allow to fetch additional metrics of interest

PerGon:
This PR is not fully ready, but I suspect it might be a bit controversial 😬 so before I go on to add tests and make it all shiny to then, possibly, not get approved, I'm opening it up in advance to gather some feedback.

This PR will add a mechanism that will allow to use other endpoints in the AWS API to fetch metrics of interest that are not exposed in Cloudwatch.

In this specific case, the need is to fetch the number of items in a DynamoDB table. This value is not exposed in Cloudwatch but it's expose in the ["describe-table" endpoint](https://docs.aws.amazon.com/cli/latest/reference/dynamodb/describe-table.html). According to the docs, this value is only updated every 6 hours (still need to add feature for not scrapping that endpoint all the time tho).
image

The result is something like this:
```
aws_dynamodb_item_count{account_id="687577111111",name="arn:aws:dynamodb:eu-west-1:687577111111:table/foobar",region="eu-west-1",tableName="foobar"} 981
aws_dynamodb_item_count{account_id="687577111111",name="arn:aws:dynamodb:eu-west-1:687577111111:table/some-table",region="eu-west-1",tableName="some-table"} 482
```

Additional IAM permissions are required.

I'm keenly aware that fetching metrics from other places that is not Cloudwatch, goes a bit against what the name of the tool says, but I decided to give it a try anyhow 🤷 

Let me know if you are willing to accept something like this. If yes, I'll work on make this PR all nice and shiny (write tests, handle errors, etc.), if not, no hard feelings 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -645,6 +645,12 @@ The following IAM permissions are required to discover tagged Database Migration
 "dms:DescribeReplicationTasks"
 ```
 
+The following IAM permissions are required to fetch additional metrics for DynamoDB (like number of items):
+```json
+"dynamodb:DescribeTable"
+```
+
+
 ## Override AWS endpoint urls
 to support local testing all AWS urls can be overridden with by setting an environment variable `AWS_ENDPOINT_URL`
 ```shell
--- a/pkg/apitagging/aws_tags.go
+++ b/pkg/apitagging/aws_tags.go
@@ -3,6 +3,7 @@ package apitagging
 import (
 	"context"
 	"errors"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbiface"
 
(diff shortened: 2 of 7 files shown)


cristiangreco:
Hey @PerGon thanks for the contribution! I understand your needs and appreciate your contribution. As you say, though, this kind of work is a bit out of scope for YACE.

I've been thinking of a plugin system to allow anybody to develop additional features for YACE, but the codebase is not stable enough at the moment to allow that.

PerGon:
Howdy!
Thanks for the feedback. I figured it was a bit of a stretch, but decided to try anyhow 😄 
A plugin system would be very nice - looking forward for that evolution of YACE.

Thanks again 🙇

PR #887, opened 2023-04-12
Refactor use of findGetMetricDataByID

cristiangreco:


pkg/job/discovery.go
```diff
@@ -94,41 +94,63 @@ func scrapeDiscoveryJobUsingMetricData(
...
+		count++
 	}
-
 	wg.Wait()
-	return resources, cw
+
+	// update getMetricDatas with fetched values and timestamps
+	for _, data := range getMetricDataOutput {
+		for _, metricDataResult := range data.MetricDataResults {
+			idx := findGetMetricDataByID(getMetricDatas, *metricDataResult.Id)
```
thepalbi, line 123: what was `metricDataResult.Id` here? can we add a comment explaining that?

pkg/job/discovery.go
```diff
@@ -94,41 +94,63 @@ func scrapeDiscoveryJobUsingMetricData(
...
+			if len(metricDataResult.Values) != 0 {
+				getMetricDatas[idx].GetMetricDataPoint = metricDataResult.Values[0]
+				getMetricDatas[idx].GetMetricDataTimestamps = metricDataResult.Timestamps[0]
+			}
+			getMetricDatas[idx].MetricID = nil // mark as processed
+		}
+	}
+
+	// remove unprocessed/unknown elements in place (if any)
+	getMetricDatas = compactSlice(getMetricDatas)
```
thepalbi, line 137: the compaction is worth because `getMetricDatas` might have a really high cardinality for example in EC2 with lots of instances + we release early memory right?

cristiangreco:
@thepalbi I've refactored the slice compaction into a separate method (think can be used in other places) and added comments. Can you please have another pass?

thepalbi (approved):
just perfect!

pkg/job/compact.go
```diff
@@ -0,0 +1,21 @@
...
+			i++
+		}
+	}
+	// nil out any left element
+	for j := i; j < len(input); j++ {
+		input[j] = nil
+	}
+	// set new slice length to allow released elements to be collected
+	return input[:i]
+}
```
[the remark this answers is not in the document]
cristiangreco: Did a benchmark before merging the PR and this showed zero allocations. Anyway I'm not entirely satisfied with the memory usage show here -- as I'm touching this part of the code in other PRs I might edit or just revert this implementation.

PR #893, opened 2023-04-14
Add counter to see if duplicate metrics are still a problem

kgeckhart:
At the end of the scraping process there's logic to remove duplicate metrics, https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/c480407b659b5fa334bb47cdc1b9ba14ceefbcfa/pkg/promutil/prometheus.go#L140-L166, which has existed for a long time. I lost the exact changeset but it was somewhere in 2018 before PR's were being done. I'm not sure it's needed anymore and the work it's doing is non-trivial.

### DIFF
--- a/pkg/exporter.go
+++ b/pkg/exporter.go
@@ -27,6 +27,7 @@ var Metrics = []prometheus.Collector{
 	promutil.Ec2APICounter,
 	promutil.DmsAPICounter,
 	promutil.StoragegatewayAPICounter,
+	promutil.DuplicateMetricsFilteredCounter,
 }
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/exporter.go
```diff
@@ -155,7 +156,8 @@ func UpdateMetrics(
 
 	metrics = append(metrics, promutil.BuildNamespaceInfoMetrics(tagsData, options.labelsSnakeCase, logger)...)
 
-	registry.MustRegister(promutil.NewPrometheusCollector(metrics))
+	collector := promutil.NewPrometheusCollector(metrics)
+	registry.MustRegister(collector)
```
kgeckhart, line 160: Given the complexity of the duplicate metric code I was surprised it didn't show up in pprof. After I made this change it started appearing. I have no idea why this is the case though.

PR #903, opened 2023-04-18
Support recentlyActive parameter in Cloudwatch API to reduce number of old metrics

PerGon:
For certain installations with a lot of instances and a lot of churn in resources (for example EC2 instances), it could be the case that old metrics are not desirable or completely irrelevant. But for the most part, they will still be present in the API if the flag `recentlyActive` is not passed.

This PR adds new configuration to a job in the form of the parameter `recentlyActiveOnly: boolean` - see [AWS docs for cloudwatch](https://docs.aws.amazon.com/cli/latest/reference/cloudwatch/list-metrics.html#options) (scroll for `recentlyActive` parameter).

In some cases the amount of data dropped can be very significant. Example (only change is flipping the `recentlyActiveOnly` flag):

#### recentlyActiveOnly is false
```
$ curl localhost:5001/metrics | wc -l
  528826
  ```
  
#### recentlyActiveOnly is true
```
$ curl localhost:5001/metrics | wc -l
  34028
  ```

Is this addition something you can consider including in YACE?
Thanks

### DIFF
--- a/pkg/clients/cloudwatch/client.go
+++ b/pkg/clients/cloudwatch/client.go
@@ -43,7 +43,7 @@ func NewClient(logger logging.Logger, cloudwatchAPI cloudwatchiface.CloudWatchAP
 	}
 }
 
-func GetListMetricsInput(metricName string, namespace string, recentlyActiveOnly bool) *cloudwatch.ListMetricsInput {
+func getListMetricsInput(metricName string, namespace string, recentlyActiveOnly bool) *cloudwatch.ListMetricsInput {
 	if !recentlyActiveOnly {
 		return &cloudwatch.ListMetricsInput{
 			MetricName: aws.String(metricName),
@@ -59,7 +59,7 @@ func GetListMetricsInput(metricName string, namespace string, recentlyActiveOnly
 }
 
 func (c client) ListMetrics(ctx context.Context, namespace string, metric *config.Metric, recentlyActiveOnly bool, fn func(page []*model.Metric)) ([]*model.Metric, error) {
-	filter := GetListMetricsInput(metric.Name, namespace, recentlyActiveOnly)
+	filter := getListMetricsInput(metric.Name, namespace, recentlyActiveOnly)
 
 	if c.logger.IsDebugEnabled() {
 		c.logger.Debug("ListMetrics", "input", filter)
--- a/pkg/clients/cloudwatch/client.go
+++ b/pkg/clients/cloudwatch/client.go
@@ -2,11 +2,13 @@ package cloudwatch
 
 import (
 	"context"
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
 	"github.com/aws/aws-sdk-go/service/cloudwatch/cloudwatchiface"
+
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/config"
-	"time"
 
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/logging"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/model"
--- a/pkg/clients/cloudwatch/client.go
+++ b/pkg/clients/cloudwatch/client.go
@@ -43,21 +43,6 @@ func NewClient(logger logging.Logger, cloudwatchAPI cloudwatchiface.CloudWatchAP
 	}
 }
 
-func getListMetricsInput(metricName string, namespace string, recentlyActiveOnly bool) *cloudwatch.ListMetricsInput {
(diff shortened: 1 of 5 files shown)


PerGon:
We have rolled out this internally, and we see HUGE improvements in CPU and Memory usage. Patched version was rolled out around 9h45 in the image below timeline.
image

If you see this coming into YACE, I'll proceed with writing the docs and what-not.

One thing to consider is that the flag `recentlyActiveOnly` could be true by default. As I see it, the vast majority of cases YACE is pulling recentish data (I might be wrong 😅 ) - so cost and performance wise, this flag default as true would make sense. What do you think?

pkg/apicloudwatch/client.go
```diff
@@ -13,7 +12,7 @@ import (
 )
 
 type CloudWatchClient interface {
-	ListMetrics(ctx context.Context, namespace string, metric *config.Metric) (*cloudwatch.ListMetricsOutput, error)
+	ListMetrics(ctx context.Context, namespace string, metric *config.Metric, recentlyActiveOnly bool) (*cloudwatch.ListMetricsOutput, error)
```
cristiangreco, line 15: @PerGon if you rebase you'll have a conflict here as the signature of this method has changed on `master`.

In general, I like that you introduced a `getListMetricsInput` func. I think we should take this opportunity to:
- make a `GetListMetricsInput` exported func
- change stuff in `pkg/job` to call it
- change the signature of `ListMetrics` to take `ctx` and `input`

cristiangreco: @PerGon do you have time to pick this up again?

PerGon: I've rebased and if I understood correctly, I got your suggestions in.

How does it look now?

cristiangreco: @PerGon the change looks good, but after merging https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/929 we're going in a different direction. I think you can basically go back to what was your original implementation, with the new param passed to `ListMetrics`. Sorry for the noise.

PerGon: No worries. Reverted back to the initial implementation - rebased to master.

How do you suggest to proceed?

pkg/clients/cloudwatch/client.go
```diff
@@ -42,11 +43,23 @@ func NewClient(logger logging.Logger, cloudwatchAPI cloudwatchiface.CloudWatchAP
 	}
 }
 
-func (c client) ListMetrics(ctx context.Context, namespace string, metric *config.Metric, fn func(page []*model.Metric)) ([]*model.Metric, error) {
-	filter := &cloudwatch.ListMetricsInput{
-		MetricName: aws.String(metric.Name),
-		Namespace:  aws.String(namespace),
+func getListMetricsInput(metricName string, namespace string, recentlyActiveOnly bool) *cloudwatch.ListMetricsInput {
```
cristiangreco, line 46: nit: could you please move this func to just after `ListMetrics`?

pkg/clients/cloudwatch/client.go
```diff
@@ -42,11 +43,23 @@ func NewClient(logger logging.Logger, cloudwatchAPI cloudwatchiface.CloudWatchAP
...
+		return &cloudwatch.ListMetricsInput{
+			MetricName: aws.String(metricName),
+			Namespace:  aws.String(namespace),
+		}
+	}
+	recentActiveString := "PT3H"
+	return &cloudwatch.ListMetricsInput{
+		MetricName:     aws.String(metricName),
+		Namespace:      aws.String(namespace),
+		RecentlyActive: &recentActiveString,
```
cristiangreco, line 57: Another nit: might be preferrable to keep a single return statement, like in 

```go
	var recentlyActive *string
	if recentlyActiveOnly {
		recentlyActive = aws.String("PT3H")
	}

	filter := &cloudwatch.ListMetricsInput{
		MetricName:     aws.String(metric.Name),
		Namespace:      aws.String(namespace),
		RecentlyActive: recentlyActive,
	}
```

pkg/config/config.go
```diff
@@ -44,6 +44,7 @@ type Job struct {
 	DimensionNameRequirements []string    `yaml:"dimensionNameRequirements"`
 	Metrics                   []*Metric   `yaml:"metrics"`
 	RoundingPeriod            *int64      `yaml:"roundingPeriod"`
+	RecentlyActiveOnly        bool        `yaml:"recentlyActiveOnly"`
```
cristiangreco, line 47: Can we please document this new param at `docs/configuration.md`?

PerGon:
Did the suggested changes 😄 Happy to add/change more things if needed 😃

cristiangreco (approved):

PR #905, opened 2023-04-20
Add feature flag to process ListMetrics response during pagination

cristiangreco:
Allow processing ListMetrics API results page-by-page rather than all at once, which should reduce peak memory usage at the expense of slightly higher processing time.

This is currently opt-in behind a new feature flag.

### DIFF
--- a/pkg/job/discovery.go
+++ b/pkg/job/discovery.go
@@ -82,6 +82,10 @@ func scrapeDiscoveryJobUsingMetricData(
 		return resources, cw
 	}
 
+	if len(resources) == 0 {
+		logger.Debug("No tagged resources", "region", region, "namespace", job.Type)
+	}
+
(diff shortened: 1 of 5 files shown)


pkg/job/discovery.go
```diff
@@ -82,6 +82,10 @@ func scrapeDiscoveryJobUsingMetricData(
 		return resources, cw
 	}
 
+	if len(resources) == 0 {
+		logger.Debug("No tagged resources", "region", region, "namespace", job.Type)
+	}
+
```
cristiangreco, line 88: This log line was within the loop below but it makes more sense to be here.

kgeckhart: I _kinda_ understand the original intent of saying "this metric can't be associated back to resources". But I 100% agree it's better here since it's a fact for the all metrics in the aws namespace not just one or two metrics.

PR #914, opened 2023-04-26
Add a test case for ActiveMQ

jesselang:
This test is currently failing, which matches the behavior we're experiencing.

https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/912


jesselang:
Resolved with #917.

PR #922, opened 2023-05-01
Refactor AWS sdk client usage to hide behind new ClientCache

kgeckhart:
The main goal of this PR is to stop SessionCache from returning any aws sdk defined interfaces.

As a part of this the name SessionCache stopped making sense as there's nothing session related in the interface. This caused me to rename it to ClientCache and move it to a new clients pkg where I centralized all the client implementations.

Related to: https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/921


pkg/clients/cloudwatch/client.go
```diff
@@ -26,23 +26,21 @@ type CloudWatchClient interface {
 	GetMetricStatistics(ctx context.Context, filter *cloudwatch.GetMetricStatisticsInput) []*cloudwatch.Datapoint
```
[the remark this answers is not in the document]
kgeckhart: Yep, I started to do that here but it spider webs out enough that I think splitting it off will make it more digestible.

pkg/clients/cache.go
```diff
@@ -16,85 +16,76 @@ import (
...
 
+	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/clients/account"
+	cloudwatch_client "github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/clients/cloudwatch"
+	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/clients/tagging"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/config"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/logging"
 )
 
-// SessionCache is an interface to a cache of sessions and clients for all the
+// ClientCache is an interface to a cache aws clients for all the
```
cristiangreco, line 38: Nit: what about `ClientsCache`?

pkg/clients/cache.go
```diff
@@ -168,192 +159,128 @@ func NewSessionCache(cfg config.ScrapeConf, fips bool, logger logging.Logger) Se
...
-		s.mu.Lock()
-		defer s.mu.Unlock()
-	}
-	if sess, ok := s.clients[role][*region]; ok && sess.asg != nil {
-		return sess.asg
-	}
-
-	s.clients[role][*region].asg = createASGSession(s.session, region, role, s.fips)
-	return s.clients[role][*region].asg
+func createCloudWatchClient(s *session.Session, region *string, role config.Role, fips bool, logger logging.Logger) cloudwatch_client.Client {
```
cristiangreco, line 223: nit: in other parts of the code `logger` is usually the first param, or the second in case `ctx` is passed as well.

pkg/clients/cloudwatch/client.go
```diff
@@ -1,4 +1,4 @@
-package apicloudwatch
+package cloudwatch
```
cristiangreco, line 1: To avoid the `cloudwatch_client` alias it might be good to keep `api` as a prefix or suffix, wdyt?

pkg/clients/cloudwatch/client.go
```diff
@@ -125,14 +123,12 @@ func (c Client) GetMetricStatistics(ctx context.Context, filter *cloudwatch.GetM
 	return resp.Datapoints
 }
 
-var _ CloudWatchClient = (*LimitedConcurrencyClient)(nil)
```
cristiangreco, line 128: Don't you think this might still be useful to double-check that the interface is implemented correctly?

cmd/yace/scraper.go
```diff
@@ -78,8 +78,6 @@ func (s *scraper) scrape(ctx context.Context, logger logging.Logger, cache sessi
 		cache,
 		exporter.MetricsPerQuery(metricsPerQuery),
 		exporter.LabelsSnakeCase(labelsSnakeCase),
-		exporter.CloudWatchAPIConcurrency(cloudwatchConcurrency),
-		exporter.TaggingAPIConcurrency(tagConcurrency),
```
cristiangreco, line 82: I _think_ these are still needed here?

cristiangreco (commented):
Looking good! Left just a few comments.

pkg/clients/cloudwatch/client.go
kgeckhart: This is still being done just in a different way, `NewClient` and `NewLimitedConcurrencyClient` return the `Client` interface instead of their concrete types. The underlying `client` and `limitedConcurrencyClient` implementations are now private. This forces a consumer to use the new functions to obtain a client and ensures the clients implement the interface.

pkg/clients/cloudwatch/client.go
kgeckhart: I would like to leave it as is because the `cloudwatch_client` alias is just temporary for this change. After the next step the only cloudwatch references should come from this package or the models package.

cmd/yace/scraper.go
kgeckhart: Oops, I forgot to put them back after a temporary refactor to hide the concurrency fully behind the client cache. I still think it might be a good thing to move but don't want to do it here.

pkg/clients/cache.go
kgeckhart: I went a slightly different way with this based on golint suggestion and dropped the package name prefix. Consumer usage will be `clients.Cache` vs `clients.ClientsCache`

pkg/clients/cloudwatch/client.go
cristiangreco: Makes sense!

thepalbi (approved):

cristiangreco (approved):

PR #929, opened 2023-05-05
Introduce model types to replace sdk types in cloudwatch client

kgeckhart:
This abstracts as much of the aws sdk as possible behind the clients pkg. After this change the only remaining references outside clients are calls to `aws.<type>` for pointers to primative types. This shouldn't interfere with swapping from v1 to v2.

I have a followup on this intended to simplify some of the data models further to eliminate reduce the use of pointers. There's a lot of places where pointers appear to have been propagated through the code purely because the aws sdk forced them to be used everywhere.

Related to: https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/921


cristiangreco:
LGTM in general. Wondering if eventually we can switch to v2-only and maybe go back to using sdk models again more extensively, which should avoid the copy overhead. wdyt?

kgeckhart:
> LGTM in general. Wondering if eventually we can switch to v2-only and maybe go back to using sdk models again more extensively, which should avoid the copy overhead. wdyt?

I don't see why not, hopefully the overhead is low enough we don't have to worry about it 🤞

PR #936, opened 2023-05-10
Add aws sdk v2 support

kgeckhart:
Adds full sdk v2 support behind a feature flag. The diff is rather large because I ended up blowing out the clients pkg. Previously we had,

```
/clients
/clients/account
/clients/cloudwatch
/clients/tagging
```

and now we have

```
/clients
/clients/v1
/clients/v2
/clients/account
/clients/account/v1
/clients/account/v2
/clients/cloudwatch
/clients/cloudwatch/v1
/clients/cloudwatch/v2
/clients/tagging
/clients/tagging/v1
/clients/tagging/v2
```

The non-versioned folders contain the interfaces and the v1/v2 folders contain the different sdk implementations. Open to other ideas on how to better isolate this. 

Luckily, we don't often make changes at this level so the maintenance burden of v1/v2 should hopefully be minimal until we can remove v1 all together.

Resolves: https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/921


pkg/clients/v2/cache.go
```diff
@@ -0,0 +1,383 @@
...
+
+			// existing settings
+			// TODO how to tell the difference between throttle and non-throttle errors now?
+			//	NumMaxRetries: 5
+			//	//MaxThrottleDelay and MinThrottleDelay used for throttle errors
+			//	MaxThrottleDelay: 10 * time.Second
+			//	MinThrottleDelay: 1 * time.Second
+			//	// For other errors
+			//	MaxRetryDelay: 3 * time.Second
+			//	MinRetryDelay: 1 * time.Second
```
kgeckhart, line 275: I kept this here for now because I still don't quite know how to port the existing config. I'm not sure it matters all that much though.

cmd/yace/main.go
```diff
@@ -234,7 +246,7 @@ func startScraper(c *cli.Context) error {
 		}
 
 		logger.Info("Reset clients cache")
-		cache = clients.NewClientCache(cfg, fips, logger)
+		cache = v1.NewClientCache(cfg, fips, logger)
```
cristiangreco, line 249: Should we pick v1/v2 here as well?

pkg/clients/cache.go
cristiangreco, line 1: Not sure where to put this comment, just wanted to say that it might be nice to have a short README.md in `pkg/clients` that explains why there's this (temporary 🤞 ) folder structure, and reminds to apply changes to both v1 and v2 if relevant. wdyt?

pkg/config/feature_flags.go
```diff
@@ -2,19 +2,21 @@ package config
...
 
+// MaxDimensionsAssociator is a feature flag used to toggle to the new resource association
+// algorithm for discovery jobs.
+const MaxDimensionsAssociator = "max-dimensions-associator"
+
 // ListMetricsCallback is a feature flag used to enable processing of ListMetrics API
 // results page by page.
 const ListMetricsCallback = "list-metrics-callback"
 
+const AwsSdkV2 = "aws-sdk-v2"
```
cristiangreco, line 18: Nit: just a little comment over here.

cristiangreco (commented):
Looking great, left some minor comments.

pkg/clients/tagging/v2/filters.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/storagegateway"
+	"github.com/grafana/regexp"
+
+	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/config"
+	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/model"
+	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/promutil"
+)
+
+type serviceFilter struct {
```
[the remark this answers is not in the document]
kgeckhart: Unfortunately I don't think it's possible 😢, the type definitions include the version specific client because we need access to the version specific AWS clients
https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/18d830652adea4e3c4e75d6c3ce3887a6b438f07/pkg/clients/tagging/v2/filters.go#L32-L32

I struggled to try to find a sharable interface between the clients at this level since v1 and v2 have completely different paging implementations.

cristiangreco (approved):

PR #939, opened 2023-05-15
Add metric yace_cloudwatch_getmetricdata_metrics_total

keyolk:
https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/339

I saw the issue above but seems that the conclusion is incorrect.
GetMetricData of CloudWatch API costs by the number of its returned metrics.

So it would be worth having how much data is fetched via GetMetricData.


ecktom:
Is there something missing here? Would be nice to get this merged as currently, there is no way to properly track costs

jaseemabid:
+1 to this feature. 

The description in the README.md to track costs is wrong, which we can fix after landing this PR. I'm happy to test and update the docs myself.

PR #974, opened 2023-06-01
Refactor clients.Cache -> clients.Factory

kgeckhart:
This change refactors `clients.Cache` to `clients.Factory` and removes the cache specific functions of `Refresh` and `Clear`.

As a part of using YACE as a library someone might choose to provide a custom implementation of `clients.Cache` when calling `UpdateMetrics`. I don't think the code in `UpdateMetrics` cares if the implementation of this interface actually caches them. I think it only cares that the implementation can provide the clients. This seems to lend itself more to a factory at this level with the "caching" requirement coming from YACE calling the library entrypoint. 

The only requirement for it to be a cache comes from the fact that `UpdateMetrics` is eventually responsible for calling `Refresh` and `Clear`. I don't think there's any particular reason `UpdateMetrics` needs to be responsible for this and I moved those calls wrap the call to `UpdateMetrics` instead.

I think this is technically a breaking change to the library entrypoint since any consumer providing a cache based implementation is now required to handle that before calling the entrypoint.

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -192,13 +192,13 @@ func startScraper(c *cli.Context) error {
 	featureFlags := c.StringSlice(enableFeatureFlag)
 
 	s := NewScraper(featureFlags)
-	cache := v1.NewClientCache(cfg, fips, logger)
+	var cache cachingFactory = v1.NewFactory(cfg, fips, logger)
 	for _, featureFlag := range featureFlags {
 		if featureFlag == config.AwsSdkV2 {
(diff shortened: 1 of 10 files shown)


thomaspeitz:
Could you reolve the conflicts and then I would merge it!

Thanks for the contribution @kgeckhart

kgeckhart:
@thomaspeitz I rebased so I think it's good to go now, thank you!

cristiangreco (approved):

Issue #816, opened 2023-02-17
[BUG] <title>YACE is not pulling any metrics for certain services when using "dimensionNameRequirements" for filtering


jordan67803:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I add "dimensionNameRequirements" to filter the metrics to only those that contain exactly this list of dimensions to my configuration file no metrics get pulled at all, and when I remove "dimensionNameRequirements" from the configuration it pulls the metrics for those services, however I can see those exact dimensions that I specified in the metrics but yet it refuses to pull the metrics when I use "dimensionNameRequirements"
exporter-yace-config.yaml.j2

### Expected Behavior

Pulls metrics that contain exactly this list of dimensions provided.

### Steps To Reproduce

Below is configuration file: 
---
apiVersion: v1alpha1
discovery:
  exportedTagsOnMetrics:
    cloudfront:
      [
        "Test.Owner",
        "Test.ArchPath",
        "Test.CostCenter",
        "Name",
        "Test.Environment",
      ]  
    es:
      [
        "Test.Owner",
        "Test.ArchPath",
        "Test.CostCenter",
        "Name",
        "Test.Environment",
      ]
  jobs:
    - type: es
      regions: ["us-east-1"]
      addCloudwatchTimestamp: true
      dimensionNameRequirements:
        - DomainName
      metrics:
        - length: 300
          period: 60
          name: 2xx
          statistics: ["Average"]
        - length: 300
          period: 60
          name: 3xx
          statistics: ["Average"]
        - length: 300
          period: 60
          name: 4xx
          statistics: ["Average"]
        - length: 300
          period: 60
          name: 5xx
          statistics: ["Average"]
        - length: 300
          period: 60
          name: AlertingDegraded
          statistics: ["Average"]
        - length: 300
          period: 60
          name: AlertingIndexExists
          statistics: ["Average"]
        - length: 300
          period: 60
          name: AlertingIndexStatus.green
          statistics: ["Average"]
        - length: 300
          period: 60
          name: AlertingIndexStatus.red
          statistics: ["Average"]
        - length: 300
          period: 60
          name: AlertingIndexStatus.yellow
          statistics: ["Average"]
        - length: 300
          period: 60
          name: AlertingNodesNotOnSchedule
          statistics: ["Average"]
        - length: 300
          period: 60
          name: AlertingNodesOnSchedule
          statistics: ["Average"]
        - length: 300
[...]

PR #984, opened 2023-06-09
Attempt to fix issue with resource association…

nmische:
…when using dimensionNameRequirements.

Not sure this is the best solution, but it seems to work for my global accelerator use case.

Addresses: #816

### DIFF
--- a/examples/ga.yml
+++ b/examples/ga.yml
@@ -1,0 +1,28 @@
+apiVersion: v1alpha1
+discovery:
+  jobs:
+    - type: AWS/GlobalAccelerator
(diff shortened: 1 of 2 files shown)


thomaspeitz:
Can you please test the behaviour with other parts of the code (or make it confiurable?)

thomaspeitz:
Thanks for the contribution :))

cristiangreco:
Don't think the issue has anything to do with associator, so closing this PR, sorry. If you have more info available to reproduce the problem please report back to the original ticket. Thanks @nmische!

PR #1030, opened 2023-06-21
Validate should fail when no roles are configured

thepalbi:
In YACE, when the environment configured role is desired, no `roles` configuration is added in each job. The `Load` function [is responsible](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/30aeceb2324763cdd024a1311045f83a09c1df36/pkg/config/config.go#L111) for configuring in the `Roles` field an empty role struct. 

However, when using yace as library, this has to be done manually. Even more, `Validate` succeeds if no roles at all are configured, making YACE to run without even scraping anything due do that main loops iterate over all roles.

This PR makes `Validate` fail when no roles are configured (not even the empty one).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -189,6 +189,8 @@ func (j *Job) validateDiscoveryJob(jobIdx int) error {
 				return err
 			}
 		}
+	} else {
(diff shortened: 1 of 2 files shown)


thomaspeitz:
Thanks!

PR #1032, opened 2023-06-26
Add support for QLDB

alexandre-alvarengazh:
Add support for AWS QLDB metrics.


pkg/config/services.go
```diff
@@ -589,6 +589,16 @@ var SupportedServices = serviceConfigs{
...
 		Alias:     "amp",
 	},
+	{
+		Namespace: "AWS/QLDB",
+		Alias:     "qldb",
+		ResourceFilters: []*string{
+			aws.String("qldb"),
+		},
+		DimensionRegexps: []*regexp.Regexp{
+			regexp.MustCompile(":ledger/(?P<LedgerName>[^/]+)"),
```
cristiangreco, line 599: Thanks @alexandre-alvarengazh! Would you mind adding a test in the `pkg/job/maxdimassociator` package for this service? You can look at [this PR](https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/937/files) for an example of what needs to be tested. In particular, we want to make sure that `LedgerName` is captured correctly as a dimension.

alexandre-alvarengazh: Sure, will add.

cristiangreco: @alexandre-alvarengazh do you still have time to work on the test cases? Otherwise I can help with that, if you could provide some example resource ARNs.

alexandre-alvarengazh: @cristiangreco Sorry for the long delay, but unit tests added.

cristiangreco:
@alexandre-alvarengazh I updated your branch to merge it but tests and linter are failing. Could you please have a look at it?

alexandre-alvarengazh:
@cristiangreco Lint and unit test fixed.

Issue #800, opened 2023-02-07
[BUG] Cannot scrape DDOS metrics


nikitakogan1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hi, 

I am trying to scrape the metric for ddos protection (DDoSDetected) for my elastic ips. I have the DDOS protection configured in AWS side and i can see the metrics in cloud watch in the needed region.
I also set the "Name" tag in all elastic ips. 

I use the following config:

```

  - type: AWS/DDoSProtection
    regions:
    - eu-central-1
    length: {{ index .Values "extraArgs" "length" }}
    searchTags:
      - key: Name
        value: .*
    metrics:
    - name: DDoSDetected
      dimensions:
        - Name: ResourceArn
          Value: "*"
      statistics:
      - Sum
      period: {{ index .Values "extraArgs" "period" }}
      nilToZero: true

```

But in the logs i see:

{"account":"xxxxxx","arn":"","error":"resource tag list is empty","job_type":"AWS/DDoSProtection","level":"error","msg":"Account contained no tagged resource. Tags must be defined for resources to be discovered.","region":"eu-central-1","time":"2023-02-07T13:41:03Z"}
{"account":"xxxxxx","arn":"","job_type":"AWS/DDoSProtection","level":"info","msg":"No tagged resources made it through filtering","region":"eu-central-1","time":"2023-02-07T13:41:03Z"}

### Expected Behavior

I expected to be able to export the metric for all the elastic Ips with the tag: Name.(All values)

### Steps To Reproduce

1. Create static ip in AWS and add it to ddos protection plan.
2. Inspect that cloudwatch creates the metrics in its side, in the needed region.
3. Try to use the config above and you should get the same log as i get. (Hopefully not, and it is something with my configuration...)

Thanks (:

cristiangreco:
Hi @nikitakogan1, could you maybe run yace with `-debug` and attach the scrubbed output?

nikitakogan1:
@cristiangreco Thanks for the answer. Here is the additional output:

```

{"account":"****","arn":"","level":"debug","msg":"GetMetricStatistics","output":"{\n  Label: \"DDoSDetected\"\n}","region":"eu-central-1","static_job_name":"DDoSDetected","time":"2023-02-07T21:01:55Z"}
[...]

PR #1040, opened 2023-06-27
ddos protection: Discover resources outside us-east-1

kgeckhart:
AWS Shield (`AWS/DDoSProtection` namespace in cloudwatch) has some non-standard scenarios when it comes to resources and metrics in cloudwatch. Shield allows you to create "global" protections and those global resources exist in us-east-1. Metrics for the global protection also end up in us-east-1 as noted by, https://docs.aws.amazon.com/waf/latest/developerguide/monitoring-cloudwatch.html.

Shield Advanced reports metrics in the US East (N. Virginia) Region, us-east-1 for the following:
- The global services Amazon CloudFront and Amazon Route 53.
- Protection groups

Resource discovery for `AWS/DDoSProtection` currently discovers the protection groups using the ResourceFilter `"shield:protection"`. This setup allows for discovery of the global protection resources and any metrics associated with them. Attempting to pull `AWS/DDoSProtection` metrics outside of us-east-1 is not possible because resource discovery does not work outside of us-east-1. The problem with this is that `AWS/DDoSProtection` also publishes metrics for all protected resources. It's unlikely those resources all exist in us-east-1 so you cannot pull any `AWS/DDoSProtection` metrics outside of us-east-1 as noted by, #800.

After talking with @cristiangreco a bit about options we landed on using a custom `ResourceFunc` for `AWS/DDoSProtection`. The implmentation uses the `shield.ListProtections` API to discover protected resources outside of us-east-1. These resources are added to the collection of `TaggedResources`. The ARN is the protected resource arn and there's a tag for the protection arn the resource is associated with. 

This solves the initial issues as well as giving a path to associate metrics. Most of the `AWS/DDoSProtection` metrics include a Dimension, `ResourceArn`, which is joinable to the new series produced by the custom `ResourceFunc`. Using some label replace it's possible to join back to the protection info metric and see protection tags as well,

```
aws_ddosprotection_ddo_sattack_requests_per_second_average + on (name) group_left(tag_ProtectionArn) aws_ddosprotection_info{tag_ProtectionArn!=""} + on (tag_ProtectionArn) group_left(tag_<your_protection_tag_here>) label_replace(aws_ddosprotection_info{tag_ProtectionArn=""}, "tag_ProtectionArn", "$1", "name", "(.+)")
```

Ideally, we could create a more simplified `aws_ddosprotection_info` series to avoid this but YACE wasn't built with this scenario in mind. I'll make a follow up ticket to describe the challenge and possible solution as solving this initial problem is complicated enough by itself.

Resolves: #800

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,8 +207,7 @@ This permission is required to discover resources for the AWS/TransitGateway nam
 
 This permission is required to discover protected resources for the AWS/DDoSProtection namespace
 ```json
-"storagegateway:ListGateways",
-"storagegateway:ListTagsForResource"
+"shield:ListProtections"
 ```
 
 If running YACE inside an AWS EC2 instance, the exporter will automatically attempt to assume the associated IAM Role. If this is undesirable behavior turn off the use the metadata endpoint by setting the environment variable `AWS_EC2_METADATA_DISABLED=true`.
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ As a quick start, the following IAM policy can be used to grant the all permissi
         "dms:DescribeReplicationTasks",
         "ec2:DescribeTransitGatewayAttachments",
         "ec2:DescribeSpotFleetRequests",
+        "shield:ListProtections",
         "storagegateway:ListGateways",
         "storagegateway:ListTagsForResource"
       ],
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -204,6 +205,12 @@ This permission is required to discover resources for the AWS/TransitGateway nam
 "ec2:DescribeTransitGatewayAttachments"
 ```
 
+This permission is required to discover protected resources for the AWS/DDoSProtection namespace
+```json
+"storagegateway:ListGateways",
+"storagegateway:ListTagsForResource"
```
cristiangreco, line 211:
```suggestion
"shield:ListProtections"
```

pkg/clients/v1/cache.go
```diff
@@ -481,3 +484,19 @@ func createAPIGatewayV2Session(sess *session.Session, region *string, role confi
 
 	return apigatewayv2.New(sess, setSTSCreds(sess, config, role))
 }
+
+func createShieldSession(sess *session.Session, region *string, role config.Role, fips bool, isDebugEnabled bool) shieldiface.ShieldAPI {
+	maxAPIGatewayAPIRetries := 5
+	config := &aws.Config{Region: region, MaxRetries: &maxAPIGatewayAPIRetries}
```
cristiangreco, line 490:
```suggestion
	maxShieldAPIRetries := 5
	config := &aws.Config{Region: region, MaxRetries: &maxShieldAPIRetries}
```

cristiangreco (approved):
🚀 🚀 🚀

PR #1089, opened 2023-08-02
Limit number of goroutines for GetMetricData calls

cristiangreco:
The CloudWatch client already limits the number of concurrent requests to GetMetricData API via the `-cloudwatch-concurrency` flag, however in discovery jobs there's no limit to how many goroutines can be spawn at the same time and get blocked waiting for the client to be available.

With this change, we use an errgroup to block adding goroutines to handle GetMetricData partitions until the client is actually available.

### DIFF
--- a/pkg/job/discovery.go
+++ b/pkg/job/discovery.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"sync"
 
+	"golang.org/x/sync/errgroup"
+
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/clients/cloudwatch"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/clients/tagging"
 	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/config"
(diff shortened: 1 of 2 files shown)


pkg/job/discovery.go
```diff
@@ -61,33 +64,46 @@ func runDiscoveryJob(
...
+		g.Go(func() error {
+			logger.Debug("GetMetricData partition", "start", start, "end", end, "partitionNum", partitionNum)
+
+			input := getMetricDatas[start:end]
+			data := clientCloudwatch.GetMetricData(gCtx, logger, input, svc.Namespace, length, job.Delay, job.RoundingPeriod)
 			if data != nil {
-				getMetricDataOutput[n] = data
+				mu.Lock()
+				getMetricDataOutput = append(getMetricDataOutput, data)
+				mu.Unlock()
```
kgeckhart, line 94: Any reason to move to locking now vs index based? It's probably safer/easier to reason about regardless just wondering

cristiangreco: Yes and another reason was to remove the preallocation of the slice (size of `partitionSize`).

PR #1106, opened 2023-08-14
move duplicated fields from CloudwatchData to a new JobContext

kgeckhart:
As a part of cleaning up some code trying to work towards https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1094 I noticed three fields, Region, AccountID, and CustomTags, are included in each CloudwatchData result instance but have the same value per job run. This PR pulls those values out to a new `JobContext` and includes all on a `CloudwatchMetricResult`. 

This means we can no longer create one large concatenated slice of `CloudwatchData` as a return to `ScrapeAwsData` because we need to keep the corresponding `JobContext` next to the slice it represents. Based on the benchmark I included,
```
goos: darwin
goarch: arm64
pkg: github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/job
BenchmarkSlices/1_job_medium_results_slice_of_slices-8          42289800                27.35 ns/op           24 B/op          1 allocs/op
BenchmarkSlices/1_job_medium_results_slice_concat-8                 4719            218349 ns/op          802820 B/op          1 allocs/op
BenchmarkSlices/1_job_small_results_slice_of_slices-8           44212057                27.28 ns/op           24 B/op          1 allocs/op
BenchmarkSlices/1_job_small_results_slice_concat-8                780162              1518 ns/op            8192 B/op          1 allocs/op
BenchmarkSlices/multiple_jobs_large_results_slice_of_slices-8    6939736               173.6 ns/op           360 B/op          4 allocs/op
BenchmarkSlices/multiple_jobs_large_results_slice_concat-8           229           5108282 ns/op        13975552 B/op          5 allocs/op
BenchmarkSlices/multiple_jobs_medium_results_slice_of_slices-8   7071829               172.3 ns/op           360 B/op          4 allocs/op
BenchmarkSlices/multiple_jobs_medium_results_slice_concat-8         3334            347094 ns/op         1327110 B/op          5 allocs/op
BenchmarkSlices/multiple_jobs_small_results_slice_of_slices-8    6845092               169.4 ns/op           360 B/op          4 allocs/op
BenchmarkSlices/multiple_jobs_small_results_slice_concat-8        817285              1328 ns/op            6784 B/op          3 allocs/op
BenchmarkSlices/1_job_large_results_slice_of_slices-8           43185079                27.44 ns/op           24 B/op          1 allocs/op
BenchmarkSlices/1_job_large_results_slice_concat-8                   633           1930388 ns/op         4005898 B/op          1 allocs/op
```
the net impact of no longer concatenating the slices is less CPU to copy from one slice to another.

I applied this same refactor for how we return `TaggedResource` from `ScrapeAwsData` as well since the net impact to the code which uses the results from `ScrapeAwsData` was really minimal.


pkg/job/scrape_test.go
kgeckhart, line 1: I can delete this before the PR is merged. I just wanted to show how I got the benchmark results

pkg/job/scrape_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+		b.Run(name+" slice of slices", func(b *testing.B) {
+			doSliceOfSlicesBench(b, input)
+		})
+
+		b.Run(name+" slice concat", func(b *testing.B) {
+			doSliceConcatBench(b, input)
+		})
+	}
+}
```
kgeckhart, line 53: I don't think this needs to stay just wanted to show the code behind the benchmarks in case I did something that would invalidate the results.

cristiangreco (approved):
🚀

PR #1107, opened 2023-08-15
Implement new concurrency limit mechanism for the CloudWatch client

thepalbi:
This PR implements a new concurrency mechanism for the CloudWatch client. For this, the client is decoupled of how the concurrency is limited, and two implementations are provided:
1. a single limit limiter, which is the current implementation
2. one that keeps one limit per api call

The second one will allow us to take full advantage of the rate limits imposed by AWS into each api.

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -134,7 +134,7 @@ func NewYACEApp() *cli.App {
 		&cli.BoolFlag{
 			Name:        "cloudwatch-concurrency.per-api-enabled",
 			Value:       exporter.DefaultCloudwatchConcurrency.PerAPIEnabled,
-			Usage:       "Whether to enable the per API CloudWatch concurrency limiter.",
+			Usage:       "Whether to enable the per API CloudWatch concurrency limiter. When enabled, the concurrency `-cloudwatch-concurrency` flag will be ignored.",
 			Destination: &cloudwatchConcurrency.PerAPIEnabled,
 		},
 		&cli.IntFlag{
@@ -232,6 +232,11 @@ func NewYACEApp() *cli.App {
 
 func startScraper(c *cli.Context) error {
 	logger = logging.NewLogger(logFormat, debug, "version", version)
+
+	// log warning if the two concurrency limiting methods are configured via CLI
+	if c.IsSet("cloudwatch-concurrency") && c.IsSet("cloudwatch-concurrency.per-api-enabled") {
+		logger.Warn("Both `cloudwatch-concurrency` and `cloudwatch-concurrency.per-api-enabled` are set. `cloudwatch-concurrency` will be ignored, and the per-api concurrency limiting strategy will be favoured.")
+	}
 
 	logger.Info("Parsing config")
 	if err := cfg.Load(configFile, logger); err != nil {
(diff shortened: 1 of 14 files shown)


cmd/yace/main.go
```diff
@@ -118,9 +119,33 @@ func NewYACEApp() *cli.App {
...
 		&cli.IntFlag{
 			Name:        "cloudwatch-concurrency",
-			Value:       exporter.DefaultCloudWatchAPIConcurrency,
+			Value:       exporter.DefaultCloudwatchConcurrency.SingleLimit,
 			Usage:       "Maximum number of concurrent requests to CloudWatch API.",
-			Destination: &cloudwatchConcurrency,
+			Destination: &cloudwatchConcurrency.SingleLimit,
+		},
+		&cli.BoolFlag{
+			Name:        "cloudwatch-concurrency.per-api-enabled",
```
cristiangreco, line 127: @thepalbi wdyt about emitting a warning if the old and new CLI flags are used together?

e.g. setting both `-cloudwatch-concurrency` and `cloudwatch-concurrency.per-api-enabled` should log a warning message.

cristiangreco: Should also be documented that  `-cloudwatch-concurrency.per-api-enabled=true` ignores whatever value you pass to `-cloudwatch-concurrency`.

pkg/clients/cloudwatch/concurrency.go
```diff
@@ -0,0 +1,69 @@
...
+	if cfg.PerAPIEnabled {
+		return NewPerAPICallLimiter(cfg.ListMetrics, cfg.GetMetricData, cfg.GetMetricStatistics)
+	}
+	return NewSingleLimiter(cfg.SingleLimit)
+}
+
+// perAPICallLimiter is a ConcurrencyLimiter that keeps a different concurrency limiter per different API call. This allows
+// a more granular control of concurrency, allowing us to take advantage of different api limits. For example, ListMetrics
+// has a limit of 25 TPS, while GetMetricData has none.
+type perAPICallLimiter map[string]chan struct{}
```
cristiangreco, line 34: Could we potentially use three different `singleLimiter` instead of the `perAPICallLimiter`? I think that would be a bit more readable perhaps?

thepalbi: yes, completely

thepalbi: ok so extracted the common funcionality into a simple semaphore type, and re-used that everywhere.

thepalbi:
@cristiangreco what do you think of the latest changes here?

pkg/clients/cloudwatch/concurrency.go
```diff
@@ -0,0 +1,105 @@
+package cloudwatch
+
+// ConcurrencyConfig configures how concurrency should be limited in a Cloudwatch API client. It allows
+// one to pick between different limiter implementations: a single limit limiter, or one with a different limit per
+// API call.
+type ConcurrencyConfig struct {
+	// PerAPIEnabled configures wether to have a limit per API call.
```
cristiangreco, line 7:
```suggestion
	// PerAPIEnabled configures whether to have a limit per API call.
```

cristiangreco (approved):

PR #1110, opened 2023-08-17
Delegate FIPS endpoint resolution to the sdk

ashrayjain:
Alternate implementation for https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1108

### DIFF
--- a/pkg/clients/v1/factory.go
+++ b/pkg/clients/v1/factory.go
@@ -1,7 +1,6 @@
 package v1
 
 import (
-	"fmt"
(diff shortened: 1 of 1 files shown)


thepalbi:
Hey @ashrayjain , thanks for taking a different stab at this. I actually like your proposal, since internally the sdk uses the same resolver I used in https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1108. Will take a look if we can add some light testing for this change in YACE

PR #1115, opened 2023-08-21
sdk v2: Set RetryMaxAttempts on root config instead client options

kgeckhart:
We have noticed that when using the various client `options.RetryMaxAttempts` it doesn't set the retry max attempts and the default of 3 retries is still being used. I opened a bug about this upstream, https://github.com/aws/aws-sdk-go-v2/issues/2251, but for now setting the retry value at the root config has worked as expected.

### DIFF
--- a/pkg/clients/v2/factory.go
+++ b/pkg/clients/v2/factory.go
@@ -90,6 +90,8 @@ func NewFactory(cfg config.ScrapeConf, fips bool, logger logging.Logger) (*Cachi
 		options = append(options, aws_config.WithUseFIPSEndpoint(aws.FIPSEndpointStateEnabled))
 	}
 
+	options = append(options, aws_config.WithRetryMaxAttempts(5))
+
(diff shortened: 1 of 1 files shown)


pkg/clients/v2/factory.go
```diff
@@ -272,6 +274,8 @@ func (c *CachingFactory) createCloudwatchClient(regionConfig *aws.Config) *cloud
 		if c.logger.IsDebugEnabled() {
 			options.ClientLogMode = aws.LogRequestWithBody | aws.LogResponseWithBody
 		}
+
+		// Setting an explicit retryer will override the default settings on the config
 		options.Retryer = retry.NewStandard(func(options *retry.StandardOptions) {
 			options.MaxAttempts = 5
 			options.MaxBackoff = 3 * time.Second
```
kgeckhart, line 281: I believe this is not impacted by the upstream bug since it's using `Retryer` instead of `RetryMaxAttempts` but it's really hard to test if that's true.

cristiangreco (approved):

PR #1130, opened 2023-09-01
scrape: Return resources even if there are no metrics

iainlane:
I was investigating why we don't see any `_info` metrics for resources in some of our regions, so dashboards which try to show "all resources" were wrong. I noticed it was correlated with seeing "No metrics data found" in the logs. That led me to discover that there are indeed no CloudWatch metrics for these resources in the affected regions. I'm guessing a bit but I think it's probably because the resources haven't generated any metrics yet (they have been unused), which means CloudWatch doesn't report them.

But I think it's still useful to know they exist in your estate, so here I'm proposing relaxing a check and keeping track of discovered resources to emit `_info` even if they don't have any metrics.

What do you think?

Regarding testing: The function I'm modifying is `ScrapeAWSData`, a public function which isn't currently tested. If you think what I'm proposing here is acceptable, we can do a little refactor to break out the inner logic and then I can write some unit tests for it.

# Commit message

If no samples have been recorded for a metric in a region, for example if the resource is new and hasn't been used yet, CloudWatch will not return the metrics via `ListMetrics`. In that case, we discard any resources we discovered and don't emit `_info` metrics for them.

But it might still be useful to know your entire estate even if some of the resources don't have metrics yet. So let's still emit the resources even when there are no metrics for them.


pabloajz (approved):

iainlane:
Based on feedback I had from @cristiangreco I'm going to update this to add a feature flag

PR #1132, opened 2023-09-01
aws sdk v2 use EndpointResolverV2

kgeckhart:
Pulls in https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1126 + necessary changes to get rid of linter and test errors.

The main difference with [EndpointResolverV2](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/#v2-endpointresolverv2--baseendpoint) is that endpoint resolution happens right before API calls are made. This moves all Endpoint related configuration off of the aws config and on to the api client options. 

The docs seem to indicate that the V2 resolver has work to do per request which v1 did not do so it's possible this has a bit of a performance penalty.


pkg/clients/v2/factory_test.go
```diff
@@ -406,19 +390,7 @@ func TestClientCache_createAutoScalingClient(t *testing.T) {
...
-
-	t.Run("Can resolve govcloud urls", func(t *testing.T) {
-		endpoint, err := options.EndpointResolver.ResolveEndpoint("us-gov-east-1", options.EndpointOptions)
-		assert.NoError(t, err)
-		assert.Equal(t, "https://autoscaling.us-gov-east-1.amazonaws.com", endpoint.URL)
-
-		endpoint, err = options.EndpointResolver.ResolveEndpoint("us-gov-west-1", options.EndpointOptions)
-		assert.NoError(t, err)
-		assert.Equal(t, "https://autoscaling.us-gov-west-1.amazonaws.com", endpoint.URL)
-	})
```
kgeckhart, line 421: This is no longer easily testable since endpoint resolution is deferred until call time. EndpointResolveV2 resolve function has a different signature `ResolveEndpoint(ctx context.Context, params EndpointParameters)` where `EndpointParameters` explicitly controls FIPS/other params. These are all mapped as a part of each API call vs the old signature.

PR #1161, opened 2023-09-20
Decouple config models from internal models

cristiangreco:
With this change, models used for config parsing are fully decoupled from models used internally at scrape time. Once the config is parsed and validated, it is converted to a representation used internally for the actual scrape logic.

The down side of this approach:
- structs appear duplicated between the `config` and `models`
- after parsing we go through a complete mapping of fields to the internal models representation (this has unnoticeable impact at startup and zero impact at runtime)

On the pros side of things:
- allows to continue improving the design of internal models without impacting the configuration
- allows multiple config format to co-exist

Coincidentally this also fixes https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1160

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -21,8 +21,15 @@ type ScrapeConf struct {
 }
 
 type Discovery struct {
-	ExportedTagsOnMetrics model.ExportedTagsOnMetrics `yaml:"exportedTagsOnMetrics"`
-	Jobs                  []*Job                      `yaml:"jobs"`
+	ExportedTagsOnMetrics ExportedTagsOnMetrics `yaml:"exportedTagsOnMetrics"`
+	Jobs                  []*Job                `yaml:"jobs"`
+}
+
+type ExportedTagsOnMetrics map[string][]string
+
+type Tag struct {
+	Key   string `yaml:"key"`
+	Value string `yaml:"value"`
 }
 
 type JobLevelMetricFields struct {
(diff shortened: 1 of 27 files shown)


pkg/config/config.go
```diff
@@ -347,6 +355,160 @@ func (m *Metric) validateMetric(metricIdx int, parent string, discovery *JobLeve
 	return nil
 }
 
+func (c *ScrapeConf) toModelConfig() model.JobsConfig {
```
kgeckhart, line 358: Since YACE has the two entry points there is a risk that this mapping diverges from what's in use in `exporter.go`. IE a field is added to `model.JobsConfig` but not given a config field/mapped from the config. I can't say I know how to easily test for something like this but just something to note.

pkg/config/config.go
```diff
@@ -347,6 +355,160 @@ func (m *Metric) validateMetric(metricIdx int, parent string, discovery *JobLeve
...
+			job.Metrics = append(job.Metrics, &model.MetricConfig{
+				Name:                   metric.Name,
+				Statistics:             metric.Statistics,
+				Period:                 metric.Period,
+				Length:                 metric.Length,
+				Delay:                  metric.Delay,
+				NilToZero:              metric.NilToZero,
+				AddCloudwatchTimestamp: metric.AddCloudwatchTimestamp,
+			})
+		}
```
kgeckhart, line 504: If I'm reading this right it looks like these three mappings show up for each job type. Thoughts on a making a `toModelRole`, `toModelTag`, and `toModelMetricConfig`?

pkg/config/config.go
```diff
@@ -347,6 +355,160 @@ func (m *Metric) validateMetric(metricIdx int, parent string, discovery *JobLeve
...
+
+		job.ExportedTagsOnMetrics = []string{}
+		if len(c.Discovery.ExportedTagsOnMetrics) > 0 {
+			svc := SupportedServices.GetService(job.Type)
+			if exportedTags, ok := c.Discovery.ExportedTagsOnMetrics[svc.Namespace]; ok {
+				job.ExportedTagsOnMetrics = exportedTags
+			} else if exportedTags, ok := c.Discovery.ExportedTagsOnMetrics[svc.Alias]; ok {
+				job.ExportedTagsOnMetrics = exportedTags
+			}
+		}
```
kgeckhart, line 384: This is the code which fixes https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1160 right? 

I'm a bit surprised that this field exists at the root of discovery (I think you mentioned this elsewhere as well). I guess you could have multiple discovery jobs for the same service but that seems more like an exception vs the normal.

kgeckhart (approved):
Just a suggestion and two small comments. Nice work going through all of these!

pkg/config/config.go
cristiangreco: I think I see what you mean. It should be responsibility of config loader to make sure that any field in the config file is mapped correctly/fully to the `model.JobsConfig`. Likewise, when adding or editing fields in `model.JobsConfig` one should go back to the config parser to populate those fields correcly (otherwise the changes might be of no use).

pkg/config/config.go
cristiangreco: > I guess you could have multiple discovery jobs for the same service but that seems more like an exception vs the normal.

Yes I think that is more of an exception -- that case however should be covered by this transformation loop.

pkg/config/config.go
cristiangreco: Good call, done!

PR #1164, opened 2023-09-25
input: fix rounding period's interaction with delay in DetermineGetMetricDataWindow

tristanburgess:
This PR adjusts the behavior of delay such that the delay is applied to the current time at the start, before the rounding period is applied. 

The intent is to create a rounding period behavior such that the rounding period always determines the boundary on which start and end timestamps can fall.

With a rounding period of 5 minutes, this means that start and end timestamps will always fall on multiples of 5 minutes. The delay parameter with this change becomes simply a parameter which changes how much time must pass after the latest rounding period ends before the start and end time reflects that new period instance. 

For example, with a rounding period of 5 minutes, and a delay of 1 minute, a wall clock time of `8:35:59` will return a start time of `8:25` and an end time of `8:30`, and a wall clock time of `8:36:00` will return a start time of `8:30` and an end time of `8:35`. With a delay of 0, the time range `8:30-8:35` would instead be returned as soon as the wall clock time hit `8:35`.


tristanburgess:
FWIW this code appears to have been largely unmodified since it was introduced in https://github.com/nerdswords/yet-another-cloudwatch-exporter/commit/e2edf61f5330cbec03d491ad5316158007c646eb. It has been shuffled around in a few refactorings, but the method content has remained unchanged.

tristanburgess:
closing, as with discussion with @cristiangreco and @kgeckhart, we want to keep the current behavior of applying delay after rounding period is applied to the current time.

Issue #1163, opened 2023-09-22
[BUG] SageMaker skipped metrics as resource not matched


GGonzalezGomez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### YACE version

v0.54.1

### Config file

apiVersion: v1alpha1
discovery:
  jobs:
  - type: AWS/SageMaker
    regions:
    - eu-west-1
    metrics:
    - name: ModelLatency
      statistics:
      - Max
      period: 600
      length: 86400

### Current Behavior

No ModelLatency metric is returned by yace despite having values in Cloudwatch. I also double checked through the aws cli and I was able to get the values. I've tested also with the "max-dimensions-associator" feature having the same behavior (missing metric).

Looking at the debug messages, for that metric it reports a "resource not matched" message and skips the metric. The dimensions ([EndpointName VariantName]) contains the regexpMapping.dimensions (EndpointName]) but when checking the signature I can see a "resource not matched" in the log.

If I select Invocations metric instead of ModelLatency (same config file but with Invocations as metric name) I can get metrics that do not use the EndpointName as dimension. In that case, dimensions are FeatureGroupName and OperationName, and the regexpMapping.dimensions is EndpointName. For that reason no signature is checked and the metric is not skipped returning values.

### Expected Behavior

I would expect to get the ModelLatency metrics as I get from Invocations as values exist.

GGonzalezGomez:
Hi,

I've done some tests for the signature calculation. Looks like the problem comes because the ARN is in lower case but the endpoint name may include upper case characters. Hence, the signature will be different in such cases. Find below the values for the labels used to call prom_model.LabelsToSignature in the associator and during the metric gathering.

Assoc labels: map[EndpointName:thisisafakeendpointname]
Assoc signature 11086376501396540421

AssociateMetricToResource labels map[EndpointName:ThisIsAFakeEndpointName]
AssociateMetricToResource signature 2945941876037630597

If I use a lower case endpointname:
Assoc labels: map[EndpointName:thisisafakeendpointname]
Assoc signature 11086376501396540421

AssociateMetricToResource labels map[EndpointName:thisisafakeendpointname]
AssociateMetricToResource signature 11086376501396540421
resource matched signature 11086376501396540421

Regards,

PR #1178, opened 2023-10-04
Associator buildLabelsMap to lower case EndpointName to match ARN

GGonzalezGomez:
This is related to #1163. The problem we are facing is that when the matching resource is being looked up, the associator is not finding it. The signature is calculated using the ARN endpoint name through the regex, but the ARN endpoint name is in lower case. If the endpoint name contains any upper case characters it fails as the signature value is different from the ARN one due to casing.
So, in order to fix it if the metric is from sagemaker and it uses the endpoint name, if it contains any upper case character it converts the endpoint name to lower case to calculate the signature.
In the issue some example is provided, hopefully this way it can be fixed.


pkg/job/maxdimassociator/associator.go
```diff
@@ -214,6 +216,15 @@ func buildLabelsMap(cwMetric *model.Metric, regexpMapping *dimensionsRegexpMappi
 				}
 			}
 
+			// AWS Sagemaker endpoint name may have upper case characters
+			// Resource ARN is only in lower case, hence transforming
+			// endpoint name value to be able to match the resource ARN
+			if cwMetric.Namespace == "AWS/SageMaker" && name == "EndpointName" {
+				if awsSageMakerEndpointUpper.MatchString(value) {
+					value = strings.ToLower(value)
```
cristiangreco, line 224: Wondering if we can always just convert to lowercase without the regex match?

GGonzalezGomez: To be honest I added it last minute to match the AmazonMQ one. I thought it would be more efficient to just convert to lowercase when needed but after some tests it is more efficient to just run the conversion every time.
I'll remove that check asap.

GGonzalezGomez: Regex has been removed now

PR #1187, opened 2023-10-11
aws sdk v2: ensure region is respected for all aws clients

kgeckhart:
The sdkv2 client creation code has two issues when it creates clients,

1. When not using AssumeRole auth the region is not properly set on the regional configurations
2. When using AssumeRole auth the `sts-region` config value and other client options are not respected on the sts client used to assume role

This PR fixes both issues and adds a tests to cover both use cases. The test for use case 1 currently fails on master. Since I had to re-structure some things to fix 2 I can't run the same test. 

Resolves https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1186 and resolves https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1029

### DIFF
--- a/pkg/clients/v2/factory.go
+++ b/pkg/clients/v2/factory.go
@@ -39,7 +39,7 @@ type awsRegion = string
 
 type CachingFactory struct {
 	logger              logging.Logger
-	stsRegion           string
+	stsOptions          func(*sts.Options)
 	clients             map[model.Role]map[awsRegion]*cachedClients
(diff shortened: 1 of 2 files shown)


cristiangreco (approved):

mikepea:
Thanks so much for merging this - it's a blocker for us to upgrade to latest.

Would it be possible for a new version to be cut to include it? ❤️

PR #1249, opened 2023-12-13
Optionally include context labels (account, region, customTags) on info metrics

kgeckhart:
This PR adds the ability to include some contextual info available to the metrics series on info metrics as well. The info metrics are really useful for being able to visualize the landscape of resources due to having tag data available on them. ATM it's not super easy to get region + account from them though,

1. They are embedded in to the `name` value which is the resource arn but this requires `label_replace` operations to extract them either at query time or an ahead of time relabel
2. Since the metric series have the data it's possible to join to get it but then you have to be careful to choose a metric which will always be present as some metrics are only published when they have a value

This simplifies that by making it easy to include it on the info metrics or not.


pkg/model/model.go
```diff
@@ -123,12 +124,12 @@ type Datapoint struct {
 	Timestamp *time.Time
 }
 
-type CloudwatchMetricResult struct {
-	Context *JobContext
-	Data    []*CloudwatchData
+type ScrapeResult[T CloudwatchData | TaggedResource] struct {
```
cristiangreco, line 127: A bit of nitpick: given that TaggedResource is only used for discovery jobs, might be more easily digestible to just have two separate structs here instead of using generics?

kgeckhart: Totally fair, I'll split this up to two separate structs

cristiangreco (approved):

PR #1254, opened 2023-12-23
Add serverless ElastiCache support

pkubicsek-sb:
Related issue: https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1253

- Extended existing EC service by recently announced GA of serverless version
- Added few simple tests to cover both serverless and nonserverless version.


pkubicsek-sb:
@cristiangreco, do you think that support for Serverless EC could be included in the next release?

cristiangreco (approved):

PR #1270, opened 2024-01-10
Improve usability and performance of searchTags

kgeckhart:
This PR includes some changes which I think will improve the performance and usability of searchTags.  

The current implementation of searchTags is a collection of key value pairs where the key must match an AWS tag name exactly and the value is treated as a regex which must match the tag value. All searchTag filters must pass for a resource to be included. This PR improves this workflow by,

1. Performance: Pre-compile the `value` regex's when mapping from config to model
    1. regex compilation was previously being done at compare time where each `value` regex would be compiled per resource
    1. Should result in a dramatic decrease in regex compiles when using searchTags 
1. Performance: Add an early exit in the event a searchTag value does not match
    1. Given the all must match semantics there's no need to keep matching as soon as we find a no-match
1. Usability: Add validation that the `value` field can be properly parsed as a regex
    1. As a part of the compare time compilation the error was being silently ignored
    1. Should prevent unexpected nil exceptions and replace them with config errors

Saved for future PR
1. Peformance: Use the searchTag `key`s as TagFilters in the `GetResources` AWS API call 
    1. The GetResources API allows for providing a set of TagFilters: https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters
    1. Their implementation is not a direct match to YACE's but the semantics that all filters must match is the same
    1. Their docs indicate `If you don't specify a value for a key, the response returns all resources that are tagged with that key, with any or no value.` which makes this change safe
    1. Moving some of the filtering over to AWS should dramatically improve performance


pkg/clients/tagging/v1/client.go
```diff
@@ -69,9 +69,18 @@ func (c client) GetResources(ctx context.Context, job model.DiscoveryJob, region
 
 	if len(svc.ResourceFilters) > 0 {
 		shouldHaveDiscoveredResources = true
+
+		var tagFilters []*resourcegroupstaggingapi.TagFilter = nil
+		if len(job.SearchTags) > 0 {
+			for _, st := range job.SearchTags {
+				tagFilters = append(tagFilters, &resourcegroupstaggingapi.TagFilter{Key: &st.Key})
```
thepalbi, line 76: can we add a comment here noting that the `TagFilter` without value filters only the existance of the tag in the resource? [docs](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html#resourcegrouptagging-GetResources-request-TagFilters)

pkg/model/model.go
```diff
@@ -179,10 +184,10 @@ func (r TaggedResource) FilterThroughTags(filterTags []Tag) bool {
 	for _, resourceTag := range r.Tags {
 		for _, filterTag := range filterTags {
 			if resourceTag.Key == filterTag.Key {
-				r, _ := regexp.Compile(filterTag.Value)
-				if r.MatchString(resourceTag.Value) {
-					tagMatches++
+				if !filterTag.Value.MatchString(resourceTag.Value) {
+					return false
 				}
+				tagMatches++
```
thepalbi, line 190: though subtle, can we add a note that this implcitly checking that all filtered key are present in the resource?

kgeckhart: This actually threw me off originally too and I almost did a bad refactor on it. A comment is absolutely  warranted

thepalbi: Haha yhea, had to stare at the code for a bit saying wait, this aint necessary

kgeckhart:
I removed the riskiest part of this PR, propagating labels filters to the AWS API, to include in a different PR that can be tested more thoroughly.

pkg/config/config.go
```diff
@@ -431,6 +438,19 @@ func toModelTags(tags []Tag) []model.Tag {
 	return ret
 }
 
+func toModelSearchTags(tags []Tag) []model.SearchTag {
+	ret := make([]model.SearchTag, 0, len(tags))
+	for _, t := range tags {
+		// regex is validated earlier
+		r, _ := regexp.Compile(t.Value)
```
nhinds, line 445: It might be worth using `regexp.MustCompile(t.Value)` instead of throwing the error away, in case there's a feature or refactor in the future that accidentally calls `toModelSearchTags()` without calling `Job.validateDiscoveryJob()`.

kgeckhart: Thanks for the comment! I see your point but feel a bit uncomfortable with potentially having unexpected panics in here. This mapping function already makes a lot of assumptions that validation has cleaned up a lot of data so it feels safe and the exported function ensures a consumer calls things properly.

thepalbi (approved):

PR #1280, opened 2024-01-19
Simplify associator usage

kgeckhart:
Small PR to simplify how the associator is used,

1. Explicitly use a "do nothing" version when there's no resources or dimension regex's to associate on
    - I know for sure this applies when there's no resources
    - Based on the code it seemed like no dimension regex's also would effectively do nothing
1. Use the same associator instance for all metrics of a service
    - It looks like the associator is service specific rather than metric specific `AssociateMetricToResource` takes the CW Metric to associate against
    - It appears as though after initialization via `NewAssociator`, `AssociateMetricToResource` only reads from the struct instance of `mappings` which should make it thread safe


cristiangreco (approved):
Great spot, I think this makes sense and is a great improvement!

pkg/job/discovery.go
```diff
@@ -209,6 +215,12 @@ func getMetricDataForQueries(
 	return getMetricDatas
 }
 
+type nothingToAssociateAssociator struct{}
```
cristiangreco, line 218: Nit: wdyt about `nopAssociator`?

Issue #1138, opened 2023-09-05
[BUG] Invalid timestamps on metrics with addCloudwatchTimestamp after upgrading from 0.53.0 to 0.54.1


doris-zhou:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### YACE version

0.54.1

### Config file

Configuration of the affected job:

```
apiVersion: v1alpha1
discovery:
  jobs:
    - type: firehose
      regions:
        - us-east-1
      period: 300
      length: 300
      addCloudwatchTimestamp: true
      metrics:
        - name: DeliveryToS3.Success
          statistics: [Average]
        - name: DeliveryToS3.DataFreshness
          statistics: [Maximum]
        - name: IncomingBytes
          statistics: [Sum]
```

### Current Behavior

We upgraded from version 0.53.0 to 0.54.1 with no changes in YACE config between versions. Prometheus started failing to scrape YACE with the message:

`expected timestamp or new record, got "INVALID"`

Prometheus logs include errors like:
```
ts=2023-09-05T15:11:17.245Z caller=scrape.go:1697 level=warn component="scrape manager" scrape_pool=<redacted> target=<redacted> msg="Error on ingesting samples with different value but same timestamp" num_dropped=1
```

and in a separate Prometheus instance:
```
ts=2023-09-05T17:55:01.215Z caller=scrape.go:1700 level=warn component="scrape manager" scrape_pool=<redacted> target=<redacted> msg="Error on ingesting samples that are too old or are too far into the future" num_dropped=33
ts=2023-09-05T18:03:35.763Z caller=scrape.go:1694 level=warn component="scrape manager" scrape_pool=<redacted> target=<redacted> msg="Error on ingesting out-of-order samples" num_dropped=42
```

I've checked YACE's metrics output and compared them between two versions. It seems that some metrics are getting a negative value for the timestamp.

With 0.53.0:

`aws_firehose_delivery_to_s3_success_average{<labels redacted>} 1 1693939800000`

With 0.54.1:

`aws_firehose_delivery_to_s3_success_average{<labels redacted>} 0 -62135596800000`

We have observed this only on the Firehose metrics so far. What could be causing the negative timestamp?

### Expected Behavior

I expect to be able to scrape metrics with the `addCloudwatchTimestamp` option as in previous versions of YACE.

thewebface:
[...]

PR #1288, opened 2024-01-30
Restore default behaviour of returning nil/absent metrics as NaN

### LINKED ISSUE #1242: [BUG] RDS metrics - values for non-existant DatabaseClass show up

### YACE version

Yet Another CloudWatch Exporter version v0.55.0

### Config file

```yaml
apiVersion: v1alpha1
sts-region: eu-central-1
discovery:
  jobs:
    - type: rds
      regions:
        - eu-central-1
      period: 300
      length: 300
      metrics:
        - name: CPUUtilization
          statistics: [Maximum]
        - name: DatabaseConnections
          statistics: [Sum]
        - name: FreeableMemory
          statistics: [Average]
        - name: FreeStorageSpace
          statistics: [Average]
        - name: ReadThroughput
          statistics: [Average]
        - name: WriteThroughput
          statistics: [Average]
        - name: ReadLatency
          statistics: [Maximum]
        - name: WriteLatency
          statistics: [Maximum]
        - name: ReadIOPS
          statistics: [Average]
        - name: WriteIOPS
          statistics: [Average]

```

### Current Behavior

After upgrading from `v0.51.0` to `v0.55.0`, I start to see RDS metrics for a non-existant instance. Usually, I wouldn't bother much, but at least from `v0.54.0` on, metrics like `aws_rds_free_storage_space_average` return a value of `0` - and that fires our alerts which makes this quite urgent. No one wants to be paged for an RDS instance that does not really exist ;)

Example: The correct and existing RDS instance is a `dimension_DatabaseClass="db.t3.small"`. However, we now get also a series for a `dimension_DatabaseClass="db.r5.xlarge"`. I looked in AWS everywhere, all regions, but I can't find that instance:

Auswahl_690

When I downgrade to `v0.53.0`, the series is still there, but returns `NaN` (at least that does not trigger any alerts, so that's what I chose to run for now):

```
aws_rds_free_storage_space_average{account_id="XXXXXXXXXXXX",dimension_DBInstanceIdentifier="",dimension_DatabaseClass="db.r5.xlarge",dimension_EngineName="",name="global",region="eu-central-1"} NaN
aws_rds_free_storage_space_average{account_id="XXXXXXXXXXXX",dimension_DBInstanceIdentifier="",dimension_DatabaseClass="db.t3.small",dimension_EngineName="",name="global",region="eu-central-1"} 9.2788228096e+10
```

### Expected Behavior

Don't show metric series for instances that don't exist :)

### Steps To Reproduce

1. Have an RDS instance in AWS eu-central-1
2. Run YACE v0.54.0 or newer
3. check `aws_rds_free_storage_space_average` wrt `dimension_DBInstanceIdentifier`

### Anything else?

I don't see those metrics in Cloudwatch. 
It might be that I overlooked something in AWS and there is a db.r5.xlarge hiding somewhere, but I think I would see that on the bill ;) Also, the `DatabaseClass` is the only Dimension that appears twice; the other dimensions (e.g. DBInstanceIdentifier) don't have ghost series
I thought maybe it's an old instance and the metrics are still around somehow, but since I cannot delete metrics in Cloudwatch, I still would have to deal with it in YACE, I think.

Any help, ideas, debug hints etc are welcome.

nhinds:
The default value for `MetricDataResult.Datapoint` was 0, so all metrics without datapoints were being returned as if they had a datapoint with the value 0, regardless of the `nilToZero` flag.

Making this field a pointer allows the v1 and v2 clients to return `nil` when there is no data, which eventually gets passed through to the existing code for handling `nil` datapoints and the `nilToZero` flag.

Fixes #1242

Might be related to #1138

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build:
 	go build -v -ldflags "$(GO_LDFLAGS)" -o yace ./cmd/yace
 
 test:
-	go test -v -bench=^$ -race -count=1 ./...
+	go test -v -bench=^$$ -race -count=1 ./...
 
 lint:
 	golangci-lint run -v -c .golangci.yml
--- a/pkg/clients/cloudwatch/v2/client_test.go
+++ b/pkg/clients/cloudwatch/v2/client_test.go
@@ -1,32 +1,17 @@
-package v1
+package v2
 
 import (
 	"testing"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+
 	"github.com/stretchr/testify/require"
 
 	cloudwatch_client "github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/clients/cloudwatch"
-	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/model"
 )
-
-func TestDimensionsToCliString(t *testing.T) {
-	// Setup Test
-
-	// Arrange
-	dimensions := []*model.Dimension{}
-	expected := ""
-
-	// Act
-	actual := dimensionsToCliString(dimensions)
-
-	// Assert
-	if actual != expected {
-		t.Fatalf("\nexpected: %q\nactual:  %q", expected, actual)
-	}
-}
 
 func Test_toMetricDataResult(t *testing.T) {
 	ts := time.Date(2024, time.January, 1, 0, 0, 0, 0, time.UTC)
@@ -41,16 +26,16 @@ func Test_toMetricDataResult(t *testing.T) {
 		{
 			name: "all metrics present",
 			getMetricDataOutput: cloudwatch.GetMetricDataOutput{
-				MetricDataResults: []*cloudwatch.MetricDataResult{
+				MetricDataResults: []types.MetricDataResult{
 					{
 						Id:         aws.String("metric-1"),
-						Values:     []*float64{aws.Float64(1.0), aws.Float64(2.0), aws.Float64(3.0)},
-						Timestamps: []*time.Time{aws.Time(ts.Add(10 * time.Minute)), aws.Time(ts.Add(5 * time.Minute)), aws.Time(ts)},
+						Values:     []float64{1.0, 2.0, 3.0},
+						Timestamps: []time.Time{ts.Add(10 * time.Minute), ts.Add(5 * time.Minute), ts},
 					},
 					{
 						Id:         aws.String("metric-2"),
-						Values:     []*float64{aws.Float64(2.0)},
-						Timestamps: []*time.Time{aws.Time(ts)},
+						Values:     []float64{2.0},
+						Timestamps: []time.Time{ts},
 					},
 				},
 			},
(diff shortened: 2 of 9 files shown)


benbridts:
I tested this to see if it fixes #1138, and it seems like it does:

I am using this config:

```yaml
---
apiVersion: v1alpha1
discovery:
  jobs:
    - type: AWS/DDoSProtection
      metrics:
        - name: DDoSDetected
          statistics: [ SampleCount ]
      addCloudwatchTimestamp: true
      period: 300
      length: 300
      regions:
        - eu-west-1

    - type: AWS/ApiGateway
      metrics:
        - name: 5XXError
          statistics: [ SampleCount ]
      addCloudwatchTimestamp: true
      period: 300
      length: 300
      regions:
        - eu-west-1
```

On the master branch, this give me (trimmed for clarity):

```
aws_apigateway_5_xxerror_sample_count{...} 0 -62135596800000
aws_ddosprotection_ddo_sdetected_sample_count{...} 0 -62135596800000
aws_apigateway_info{...} 0
aws_ddosprotection_info{...} 0
```

with this branch, I get:

```
aws_apigateway_info{...} 0
aws_ddosprotection_info{...} 0
```

which is the same as with previous versions like `v0.49.0` and `v0.50.0`.

If I remove `addCloudwatchTimestamp: true`, I get on master:
```
aws_ddosprotection_ddo_sdetected_sample_count{...} 0
aws_apigateway_5_xxerror_sample_count{...} 0
aws_apigateway_info{...} 0
aws_ddosprotection_info{...} 0
``` 
and on this branch:
```
aws_ddosprotection_ddo_sdetected_sample_count{...} NaN
aws_apigateway_5_xxerror_sample_count{...} NaN
aws_apigateway_info{...} 0
aws_ddosprotection_info{...} 0
```

which is the same as with previous versions like `v0.49.0`  I also got the NaN behaviour, so this PR looks fine as a fix for #1138  (in previous versions the`aws_ddosprotection_ddo_sdetected_sample_count` is missing, but that is unrelated to this PR.

nhinds:
This will be solved differently in #1289, closing

nhinds:
Linting failures seem to be new between `golangci-lint` `1.56.0` and `1.56.1` - `docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.56.0 golangci-lint run -v` succeeds, while `docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.56.1 golangci-lint run -v` fails. The same code also passed linting 2 weeks ago on version `1.55.2` (https://github.com/nerdswords/yet-another-cloudwatch-exporter/actions/runs/7705664092/job/20999963536). I assume the CI is floating on the latest version of `golangci-lint` as I don't see any commits that would bump this.

The failures seem to be in files I haven't touched (`pkg/clients/tagging/*, `pkg/clients/v1/*`, `pkg/job/compact_test.go`, `cmd/yace/*`)

Makefile
```diff
@@ -9,7 +9,7 @@ build:
 	go build -v -ldflags "$(GO_LDFLAGS)" -o yace ./cmd/yace
 
 test:
-	go test -v -bench=^$ -race -count=1 ./...
+	go test -v -bench=^$$ -race -count=1 ./...
```
cristiangreco, line 12: @nhinds what is this needed for?

cristiangreco:
> The failures seem to be in files I haven't touched (`pkg/clients/tagging/*`, `pkg/clients/v1/*`, `pkg/job/compact_test.go`, `cmd/yace/*`)

Can you please rebase? The linter errors have been fixed in https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1295

Makefile
nhinds: `$` is a special character for `make`, I noticed that in the CI logs and when running `make test` locally the command was incorrect.

From ac751ac090b2d5322c0cf1c18ca1e03714833a75:

>  Escape $ in Makefile
>
> The `$` in the Makefile was being interpreted by `make`, and running the
> incorrect command `go test -v -bench=^-race -count=1 ./...`

I can send commit as its own PR if you prefer?

cristiangreco: Great spot! No need to open a separate PR.

pkg/clients/cloudwatch/v2/client_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+
+	"github.com/stretchr/testify/require"
+
+	cloudwatch_client "github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/clients/cloudwatch"
+)
+
+func Test_toMetricDataResult(t *testing.T) {
+	ts := time.Now()
```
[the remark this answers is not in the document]
nhinds: Sure - I was matching the pattern in `TestBuildMetrics` from `pkg/promutil/migrate_test.go`, so I will change that one to use the same constant value.

cristiangreco (approved):
Great work! 🚀

PR #1289, opened 2024-02-02
Restructure fields on CloudwatchData

kgeckhart:
This PR restructures some of the fields on the `CloudwatchData` struct to make the field usage much more obvious. This struct is used through out the scraping process to carry around context and the result of finding metric data and contains a lot of overlapping/overloaded fields. The net result should be a struct that is much easier to understand making some of the complex mapping done in `migrate.BuildMetrics` a little more obvious, and less pointers where they are not necessary.

Related to: https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1290

### DIFF
--- a/pkg/clients/cloudwatch/v1/input.go
+++ b/pkg/clients/cloudwatch/v1/input.go
@@ -31,7 +31,7 @@ func createGetMetricDataInput(getMetricData []*model.CloudwatchData, namespace *
 			Stat:   &data.GetMetricDataResult.Statistic,
 		}
 		metricsDataQuery = append(metricsDataQuery, &cloudwatch.MetricDataQuery{
-			Id:         data.GetMetricDataResult.ID,
+			Id:         &data.GetMetricDataResult.ID,
 			MetricStat: metricStat,
 			ReturnData: aws.Bool(true),
 		})
--- a/pkg/clients/cloudwatch/v1/input.go
+++ b/pkg/clients/cloudwatch/v1/input.go
@@ -62,9 +62,11 @@ func createGetMetricDataInput(getMetricData []*model.CloudwatchData, namespace *
 func toCloudWatchDimensions(dimensions []model.Dimension) []*cloudwatch.Dimension {
 	cwDim := make([]*cloudwatch.Dimension, 0, len(dimensions))
 	for _, dim := range dimensions {
+		// Don't take pointers directly to loop variables
+		cDim := dim
 		cwDim = append(cwDim, &cloudwatch.Dimension{
-			Name:  &dim.Name,
-			Value: &dim.Value,
+			Name:  &cDim.Name,
+			Value: &cDim.Value,
 		})
 	}
 	return cwDim
--- a/pkg/clients/cloudwatch/v1/input.go
+++ b/pkg/clients/cloudwatch/v1/input.go
@@ -28,7 +28,7 @@ func createGetMetricDataInput(getMetricData []*model.CloudwatchData, namespace *
 				Namespace:  namespace,
 			},
 			Period: &data.GetMetricDataProcessingParams.Period,
-			Stat:   &data.GetMetricDataResult.Statistic,
+			Stat:   &data.GetMetricDataProcessingParams.Statistic,
 		}
 		metricsDataQuery = append(metricsDataQuery, &cloudwatch.MetricDataQuery{
 			Id:         &data.GetMetricDataProcessingParams.QueryID,
--- a/pkg/clients/cloudwatch/v1/input.go
+++ b/pkg/clients/cloudwatch/v1/input.go
@@ -18,20 +18,20 @@ func createGetMetricDataInput(getMetricData []*model.CloudwatchData, namespace *
 	metricsDataQuery := make([]*cloudwatch.MetricDataQuery, 0, len(getMetricData))
 	roundingPeriod := model.DefaultPeriodSeconds
 	for _, data := range getMetricData {
-		if data.MetricConfig.Period < roundingPeriod {
-			roundingPeriod = data.MetricConfig.Period
+		if data.GetMetricDataProcessingParams.Period < roundingPeriod {
+			roundingPeriod = data.GetMetricDataProcessingParams.Period
 		}
 		metricStat := &cloudwatch.MetricStat{
 			Metric: &cloudwatch.Metric{
 				Dimensions: toCloudWatchDimensions(data.Dimensions),
-				MetricName: &data.MetricConfig.Name,
+				MetricName: &data.MetricName,
 				Namespace:  namespace,
 			},
-			Period: &data.MetricConfig.Period,
+			Period: &data.GetMetricDataProcessingParams.Period,
 			Stat:   &data.GetMetricDataResult.Statistic,
 		}
 		metricsDataQuery = append(metricsDataQuery, &cloudwatch.MetricDataQuery{
-			Id:         &data.GetMetricDataResult.ID,
+			Id:         &data.GetMetricDataProcessingParams.QueryID,
 			MetricStat: metricStat,
 			ReturnData: aws.Bool(true),
 		})
--- a/pkg/clients/cloudwatch/v1/input.go
+++ b/pkg/clients/cloudwatch/v1/input.go
@@ -18,20 +18,20 @@ func createGetMetricDataInput(getMetricData []*model.CloudwatchData, namespace *
 	metricsDataQuery := make([]*cloudwatch.MetricDataQuery, 0, len(getMetricData))
 	roundingPeriod := model.DefaultPeriodSeconds
 	for _, data := range getMetricData {
-		if data.Period < roundingPeriod {
-			roundingPeriod = data.Period
+		if data.MetricConfig.Period < roundingPeriod {
+			roundingPeriod = data.MetricConfig.Period
 		}
 		metricStat := &cloudwatch.MetricStat{
 			Metric: &cloudwatch.Metric{
 				Dimensions: toCloudWatchDimensions(data.Dimensions),
-				MetricName: data.Metric,
+				MetricName: &data.MetricConfig.Name,
 				Namespace:  namespace,
 			},
-			Period: &data.Period,
-			Stat:   &data.Statistics[0],
+			Period: &data.MetricConfig.Period,
+			Stat:   &data.GetMetricDataResult.Statistic,
 		}
 		metricsDataQuery = append(metricsDataQuery, &cloudwatch.MetricDataQuery{
-			Id:         data.MetricID,
+			Id:         data.GetMetricDataResult.ID,
 			MetricStat: metricStat,
 			ReturnData: aws.Bool(true),
 		})
--- a/pkg/clients/cloudwatch/v1/input.go
+++ b/pkg/clients/cloudwatch/v1/input.go
@@ -59,7 +59,7 @@ func createGetMetricDataInput(getMetricData []*model.CloudwatchData, namespace *
 	}
 }
 
-func toCloudWatchDimensions(dimensions []*model.Dimension) []*cloudwatch.Dimension {
+func toCloudWatchDimensions(dimensions []model.Dimension) []*cloudwatch.Dimension {
 	cwDim := make([]*cloudwatch.Dimension, 0, len(dimensions))
 	for _, dim := range dimensions {
 		cwDim = append(cwDim, &cloudwatch.Dimension{
@@ -70,7 +70,7 @@ func toCloudWatchDimensions(dimensions []*model.Dimension) []*cloudwatch.Dimensi
 	return cwDim
 }
 
-func createGetMetricStatisticsInput(dimensions []*model.Dimension, namespace *string, metric *model.MetricConfig, logger logging.Logger) *cloudwatch.GetMetricStatisticsInput {
+func createGetMetricStatisticsInput(dimensions []model.Dimension, namespace *string, metric *model.MetricConfig, logger logging.Logger) *cloudwatch.GetMetricStatisticsInput {
 	period := metric.Period
 	length := metric.Length
 	delay := metric.Delay
@@ -113,7 +113,7 @@ func createGetMetricStatisticsInput(dimensions []*model.Dimension, namespace *st
 	return output
 }
 
-func dimensionsToCliString(dimensions []*model.Dimension) string {
+func dimensionsToCliString(dimensions []model.Dimension) string {
 	out := strings.Builder{}
 	for _, dim := range dimensions {
 		out.WriteString("Name=")
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -168,10 +168,14 @@ type GetMetricStatisticResult struct {
 }
 
 type GetMetricDataResult struct {
-	ID        *string
-	Statistic string
-	Datapoint float64
-	Timestamp time.Time
+	// GetMetricData Query ID used for results mapping
+	ID string
+	// Indicates if this result was successfully mapped to a query result, only used during metric gathering
+	// and does not need to be respected afterward
+	MappedToAQueryResult bool
+	Statistic            string
+	Datapoint            float64
+	Timestamp            time.Time
 }
 
 // TaggedResource is an AWS resource with tags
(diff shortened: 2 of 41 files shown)


pkg/model/model.go
```diff
@@ -149,19 +149,24 @@ type ScrapeContext struct {
...
-	Statistics              []string
-	Points                  []*Datapoint
-	GetMetricDataPoint      *float64
-	GetMetricDataTimestamps time.Time
-	NilToZero               *bool
-	AddCloudwatchTimestamp  *bool
-	Tags                    []Tag
-	Dimensions              []*Dimension
-	Period                  int64
+	ID                       string
```
kgeckhart, line 152: This is a very overloaded/poorly named field but I couldn't find a better name,

- In a discovery job this is the [resource arn](https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/c4ddf3cc6c0ba5d08510dd992443b5ea25ce7a8b/pkg/job/discovery.go#L262-L262)
- In a custom namespace job this is the [custom namespace job name](https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/c4ddf3cc6c0ba5d08510dd992443b5ea25ce7a8b/pkg/job/custom.go#L114-L114)
- In a static job this is the [resource name](https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/c4ddf3cc6c0ba5d08510dd992443b5ea25ce7a8b/pkg/job/static.go#L29-L29)

`Resource` almost works except custom namespace jobs. It eventually is put on the metric in the `name` label but `name` doesn't feel any better

pkg/model/model.go
```diff
@@ -149,19 +149,24 @@ type ScrapeContext struct {
...
+	Namespace                string
+	Tags                     []Tag
+	Dimensions               []*Dimension
+	MetricConfig             *MetricConfig
```
kgeckhart, line 156: ATM this might look like it's going to add unnecessary data. I have a follow up which is going to need all the fields to make the batching done in `runDiscoveryJob` make sure all `GetMetricData` queries use an appropriate start and endtime based on the values in metric configuration.

pkg/promutil/migrate.go
```diff
@@ -64,63 +63,76 @@ func BuildMetrics(results []model.CloudwatchMetricResult, labelsSnakeCase bool,
...
+					timestamp = ts
+					if dataPoint == nil {
+						exportedDatapoint = math.NaN()
+						// If there was no datapoint then timestamp is a default value
+						includeTimestamp = false
 					}
 				}
+				if *metric.MetricConfig.NilToZero && math.IsNaN(exportedDatapoint) {
+					exportedDatapoint = 0
+				}
```
kgeckhart, line 98: I think this change fixes a bug which exists with `nilToZero`. A discovery job can never have a `GetMetricDataResult` with a nil DataPoint. When we map the data from AWS `Datapoint` is one of the only fields [which is not a pointer](https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/c4ddf3cc6c0ba5d08510dd992443b5ea25ce7a8b/pkg/clients/cloudwatch/client.go#L45-L49). After that we [remove all metrics from the result](https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/c4ddf3cc6c0ba5d08510dd992443b5ea25ce7a8b/pkg/job/discovery.go#L109-L114) which didn't have a datapoint mapped. 

Based on this I think the old code is completely broken and could never work for a discovery job,

```go
if exportedDatapoint == nil && (metric.AddCloudwatchTimestamp == nil || !*metric.AddCloudwatchTimestamp) {
    exportedDatapoint = aws.Float64(math.NaN())
    includeTimestamp = false
    if *metric.NilToZero {
        exportedDatapoint = aws.Float64(0)
```
We only use `nilToZero` when the actual value is nil (can never happen) and some extra conditionals related if the CloudWatch timestamp is to be used. I don't understand how those two conditions are related at all?

The new logic here splits this up so that we properly handle the valid nil data point case for a static job and all jobs can have `NaN` converted to `0` according to `NilToZero`.

pkg/model/model.go
thepalbi: `ResourceName`, and a nice comment on top clarifying the three possible meanings?

pkg/promutil/migrate.go
thepalbi: yhea that logic above is really weird (like the mix of checking if there's data and the CW ts should be used). +1 on dropping that in favour of yours

pkg/promutil/migrate.go
```diff
@@ -64,63 +63,76 @@ func BuildMetrics(results []model.CloudwatchMetricResult, labelsSnakeCase bool,
...
+			if metric.GetMetricDataResult != nil {
+				statisticsRepresentedInMetric = []string{metric.GetMetricDataResult.Statistic}
+			} else {
+				statisticsRepresentedInMetric = metric.MetricConfig.Statistics
+			}
+			for _, statistic := range statisticsRepresentedInMetric {
+				var exportedDatapoint float64
+				var timestamp time.Time
+				if metric.GetMetricDataResult != nil {
+					exportedDatapoint = metric.GetMetricDataResult.Datapoint
```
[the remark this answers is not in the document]
kgeckhart: Yeah a comment here makes sense the content will be slightly different though. The way I refactored it, it's not possible at all for the value to be `nil` as it's no longer a pointer. It is possible that it is `NaN` though if the GetMetricData API returns a NaN value to us. I couldn't find if this was possible or not but if I search for `NaN` in the sdk I see that their have support for `NaN` so it _seems_ possible.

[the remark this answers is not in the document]
nhinds: There's a bug in recent versions where absent metrics are always returned at `0` instead of `nil`/`NaN`, because `MetricDataResult.Datapoint` wasn't a pointer and defaulted everything to `0`.

I attempted to address this in #1288 by making `MetricDataResult.Datapoint` a pointer and ensuring that `nil` is passed through for metrics with no values, though I see that this change makes `nil` impossible in either layer as now neither are pointers. In practice AWS does not return `NaN` datapoints for missing metrics, they return `Values: []` instead, and `pkg/clients/cloudwatch/*.toMetricDataResult()` has to check for this.

You could use `NaN` as a marker value for absent metrics if you want to avoid pointers, but `toMetricDataResult()` would need to be updated to explicitly populate that `NaN` instead of using the default value of `0` in `MetricDataResult.Datapoint`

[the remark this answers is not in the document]
cristiangreco: Agree that adding a marker in `toMetricDataResult()` is a more appropriate fix than the one proposed in https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1288. If you're all ok with it, I think the problem could be addressed in this PR while we're refactoring models.

pkg/promutil/migrate.go
cristiangreco: > A discovery job can never have a GetMetricDataResult with a nil DataPoint.

Before the change where we stopped using pointers, it could have been nil in the case the CW returns the metric without any data point (I think it's the case that @nhinds refers to in the other comment).

pkg/promutil/migrate.go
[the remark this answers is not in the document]
kgeckhart: This is awesome because it's the exact reason I wanted to refactor the shape of this struct. There's too much reliance on implicit behavior rather than explicit, in this case there was a reliance on the fact that the datapoint was default initialized to `nil` rather than explicitly being marked as "not present". My plan is to split the new struct I made `GetMetricDataResult` in to two,

```go
type GetMetricDataProcessingParams struct {
	// GetMetricData Query MetricQueryID used for results mapping
	MetricQueryID string
	// Indicates if this result was successfully mapped to a query result, only used during metric gathering
	// and does not need to be respected afterward
	MappedToAQueryResult bool
}

type GetMetricDataResult struct {
	Statistic    string
	HadDatapoint bool
	Datapoint    float64
	Timestamp    time.Time
}
```

The eventually I would drop `GetMetricDataProcessingParams` from this struct and hide that complexity behind the new GetMetricDataProcessor abstraction (https://github.com/kgeckhart/yet-another-cloudwatch-exporter/pull/2). 

I'm choosing to be super explicit with a `HadDatapoint bool` vs using `NaN` in the Datapoint so we can tell if the `Timestamp` value is usable for `AddCloudwatchTimestamp` (https://github.com/kgeckhart/yet-another-cloudwatch-exporter/blob/47cbd27471262253a832b826b1ae0c8eef216529/pkg/promutil/migrate.go#L87-L91).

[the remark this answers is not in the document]
kgeckhart: @nhinds I rebased my PR on top of yours because I didn't want to lose the great work you did to add tests where they did not exist previously. Thank you so much for doing this!

I did notice one odd thing when I pulled them over where for one of the new tests you added this metric, https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1288/files#diff-74544d5174e5eca3f8554da703851b2282e736b61cce5530e02d405b31ebbd94R453-R468, is dropped after the call to `BuildMetrics`. After I finished the rebase I had to update the test to include it in the results which I think is valid because `BuildMetrics` shouldn't be causing any metrics to be dropped. If you would be able to look at it that would be helpful.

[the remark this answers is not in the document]
nhinds: > I was hoping it would leave the changesets associated with you but it looks like it pulled your name but not your github info

...so, it turns out I committed those with my work details, which aren't tied to my personal github account. That's my fault, I was going to re-do them if I hadn't closed my PR and didn't realize you would pull them in here. Thanks for picking up the commits, I was considering sending a followup PR to add the tests once this was merged and I'm glad to not have to rewrite them.

> I would happily take my changes and apply them to your branch if you felt comfortable giving me access to push to the branch in your fork or any other option so you don't lose them.

I've updated the branch on my fork with the correct author (https://github.com/nhinds/yet-another-cloudwatch-exporter/tree/allow-nil-nan-metrics) if you wanted to rebase, but feel free to leave them as-is if that's too much effort.

> After I finished the rebase I had to update the test to include it in the results which I think is valid because BuildMetrics shouldn't be causing any metrics to be dropped. If you would be able to look at it that would be helpful.

With the code on my branch, `GetMetricDataPoint` would be `nil` if there were no datapoints returned from CloudWatch. If there are no datapoints, there are also no datapoint timestamps - I think my test data is slightly misleading there, it should have also set `GetMetricDataTimestamps: time.Time{}` rather than `GetMetricDataTimestamps: ts`.

That particular testcase checked the logic where if `AddCloudwatchTimestamp` is `true` and there is no timestamp available from Cloudwatch, the metric is skipped as we don't have a timestamp from cloudwatch.

It looks like the new behaviour in this PR will result in a metric being returned as `NaN` or `0` without a timestamp in this scenario, which I think that will cause Prometheus to interpret the `NaN` or `0` as a datapoint with a timestamp equal to the scrape time. That might cause problems if a future scrape _does_ find data from cloudwatch, as the timestamp from cloudwatch may end up being earlier than the last scrape time. I think Prometheus doesn't like ingesting metrics whose timestamps end up out-of-order.

Personally I do not use `AddCloudwatchTimestamp`, so it will not impact me either way. I don't know if this will be a practical problem for people who _do_ use it, but I would be hesitant to change the behaviour if it would be simple to continue to skip metrics without results when the user has requested timestamps be returned from cloudwatch.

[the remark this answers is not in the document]
kgeckhart: If @cristiangreco is okay with it @nhinds what I think would be easiest here is you re-open your PR and it gets merged. I really didn't intend this PR to change any logic I wanted to just move around fields and remove unnecessary pointers. I happened to stubble on to this part which looked weird while you actually fixed it + added tests.

The corner case with `AddCloudwatchTimestamp` is rather annoying but I also agree that dropping the metric is the right thing to do. More recent versions of prometheus support out of order samples but even then the data ends up being incredibly confusing to use. CloudWatch timestamps are always behind by the period amount so a non-CloudWatch timestamp is going be in line with the next CloudWatch sample but not rounded due to scrape time. I'll clean this up on my branch + add some clarifying comments. All of this + the tests will hopefully prevent a future regression like this again.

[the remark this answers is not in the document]
cristiangreco: Sorry for the late reply. Yes please, @nhinds can you re-open your PR and tag me for review?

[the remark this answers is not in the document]
nhinds: Reopened #1288

pkg/clients/cloudwatch/v1/input.go
```diff
@@ -59,18 +59,20 @@ func createGetMetricDataInput(getMetricData []*model.CloudwatchData, namespace *
 	}
 }
 
-func toCloudWatchDimensions(dimensions []*model.Dimension) []*cloudwatch.Dimension {
+func toCloudWatchDimensions(dimensions []model.Dimension) []*cloudwatch.Dimension {
 	cwDim := make([]*cloudwatch.Dimension, 0, len(dimensions))
 	for _, dim := range dimensions {
+		// Don't take pointers directly to loop variables
+		cDim := dim
```
nhinds, line 66: The code below doesn't take a pointer to a loop variable, it takes a pointer to a field in a struct, which behaves the same with & without this `cDim` copy: https://go.dev/play/p/WUzQblUn-2z. I removed a couple of similar copies in !1288 which turned out to have been unnecessary before & after the change.

It may be easier to prove that it's safe if YACE stops using explicit pointers and uses the AWS SDK's pointer conversion utilities ([v1](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.50.20/aws#hdr-Value_and_Pointer_Conversion_Utilities), [v2](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v1.25.0/aws#hdr-Value_and_Pointer_Conversion_Utilities)), e.g. `aws.String(dim.Name)` and `aws.String(dim.Value)`. Those helpers are given a string as a value/copy, so they never need callers to make their own copies of the string variables.

nhinds: Apologies, this is my mistake. This behaviour changed in 1.22 (the default for go.dev/play), and the copy _is_ required in 1.21 and older for this pattern. Switching the dropdown to 1.21 on go.dev/play shows the wrong results.

Using `aws.String()` would avoid the copy in both versions, but isn't necessary if you use `cDim` here.

cristiangreco: Seems that the next golangci release should include some tooling to check all the unnecessary loop variables, I'll check on that once it's released.

kgeckhart: I'm sooooo happy they fixed this behavior in 1.22. If there wasn't a linter to catch it I would get burned by it constantly.

kgeckhart:
@thepalbi, @nhinds, @cristiangreco I got this back in to a spot where I'm incredibly happy with the result of the changes. Let me know what you think.

pkg/promutil/migrate.go
```diff
@@ -64,63 +63,89 @@ func BuildMetrics(results []model.CloudwatchMetricResult, labelsSnakeCase bool,
...
 				}
-				if exportedDatapoint == nil && (metric.AddCloudwatchTimestamp == nil || !*metric.AddCloudwatchTimestamp) {
-					exportedDatapoint = aws.Float64(math.NaN())
-					includeTimestamp = false
-					if *metric.NilToZero {
-						exportedDatapoint = aws.Float64(0)
-					}
+				var exportedDatapoint float64
+				if dataPoint == nil && metric.MetricMigrationParams.AddCloudwatchTimestamp {
+					// If we did not get a datapoint then the is default making it unusable in the exported metric
```
nhinds, line 78: I don't understand this line. Is there a typo near "then the is default making it unusable"? Should it say something like "there is no default timestamp" or "then the timestamp is also missing"?

nhinds (approved):

pkg/promutil/migrate.go
kgeckhart: 100% a typo, I'll reword it

nhinds (approved):

pkg/model/model.go
```diff
@@ -154,19 +154,55 @@ type ScrapeContext struct {
...
+}
+
+type GetMetricDataProcessingParams struct {
+	// QueryID is the query ID for GetMetricData used in results mapping
+	QueryID string
+
+	// The statistic to be used to call GetMetricData
+	Statistic string
+
+	// Fields which impact the start and endtime for
```
cristiangreco, line 191: Nit: can you review this comment? (in a followup)

cristiangreco (approved):
🚀

PR #1306, opened 2024-02-15
Add support for KMS metrics.

daharon:
I added support for Key Management Service metrics.

Example metrics:
```
aws_kms_seconds_until_key_material_expiration_maximum{account_id="****",dimension_KeyId="65266e81-b1ac-437a-bba1-dd82e2f7f7ee",name="arn:aws:kms:us-east-2:****:key/65266e81-b1ac-437a-bba1-dd82e2f7f7ee",region="us-east-2"} 2.8925349e+07
```


pkg/config/services.go
```diff
@@ -512,6 +512,16 @@ var SupportedServices = serviceConfigs{
...
+	{
+		Namespace: "AWS/KMS",
+		Alias:     "kms",
+		ResourceFilters: []*string{
+			aws.String("kms:key"),
+		},
+		DimensionRegexps: []*regexp.Regexp{
+			regexp.MustCompile(":key/(?P<KeyId>[^/]+)"),
+		},
+	},
```
cristiangreco, line 524: Can you please also add a test case for the associator?

daharon: Added the test case.

cristiangreco (approved):

PR #1325, opened 2024-03-04
Add abstraction for GetMetricsData processing

kgeckhart:
This introduces a new package + abstraction `getmetricsdata.processor`which takes the responsibility of batching the requests, calling GetMetricData, and mapping the results. Now that an abstraction is in place we can share the logic between discovery jobs and custom jobs reducing the amount of code copy between the two and increase test coverage on this code path 🥳 .

This PR is intended to contain no logic changes just reorganization of code + adding tests.

Related to: https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1290
Related to: https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1094


cristiangreco (approved):
🚀

Issue #1324, opened 2024-03-03
[FEATURE] Add AWS/EC2CapacityReservations Metrics Support


luismy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Feature description

AWS supports CloudWatch metrics for `AWS/EC2CapacityReservations` according to their [docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservation-cw-metrics.html)

I believe this service is quite straightforward as there are only 4 metrics with the ability to refine by `CapacityReservationId` as dimension

### What might the configuration look like?

```
- type: AWS/EC2CapacityReservations
  regions:
    - ew-west-2
  metrics:
    - name: UsedInstanceCount
      statistics: [Average]
    - name: AvailableInstanceCount
      statistics: [Average]
    - name: TotalInstanceCount
      statistics: [Average]
    - name: InstanceUtilization
      statistics: [Average]
```

### Anything else?

AWS documentation at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservation-cw-metrics.html

I'm happy to give this a go if somebody can point out how new metrics are added to be supported in YACE. I'll look into the code anyway but any pointers will be great

PR #1341, opened 2024-03-16
Add AWS/EC2CapacityReservations to the services list

luismy:
## Summary
Add `AWS/EC2CapacityReservations` to the services list

## Context
See issue #1324


luismy:
@cristiangreco would you mind reviewing this one? Not sure if I also need to include a test and/or update docs. Happy to do that if required

pkg/config/services.go
```diff
@@ -334,6 +334,10 @@ var SupportedServices = serviceConfigs{
 			regexp.MustCompile("(?P<FleetRequestId>.*)"),
 		},
 	},
+	{
+		Namespace: "AWS/EC2CapacityReservations",
+		Alias:     "ec2CapacityReservations",
```
cristiangreco, line 339:
```suggestion
		Alias:     "ec2CapacityReservations",
		DimensionRegexps: []*regexp.Regexp{
				regexp.MustCompile(":capacity-reservation/(?P<CapacityReservationId>)$"),
		},
```

cristiangreco: I haven't tested this... do need to define a regexp here in order to capture tagged resources?

luismy:
I've tested against my sandbox AWS account and you get all the reservation
ids if I remember correctly. I can add that for completeness as it will be
optional

Let me run a couple of tests against my account and I will let you know

cristiangreco:
@luismy any update on this one?

pkg/config/services.go
kgeckhart: According to https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html that's would match the format for a capacity-reservation resource
image

cristiangreco: Yep, I initially drafted it following that spec. I'll go ahead and merge the PR with this change included, we can always fix later if necessary.

PR #1359, opened 2024-04-03
Support partner events buses

HristoStoyanovYotpo:
AWS/Events partner buses are currently not fetched by YACE, because their names contain / character.
This is written here on the EventBus Name property: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-eventbus.html
```
Name
The name of the new event bus.

Custom event bus names can't contain the / character, but you can use the / character in partner event bus names. In addition, for partner event buses, the name must exactly match the name of the partner event source that this event bus is matched to.
```

This PR is expected to make metrics for buses with names containing / character to also be captured by YACE

### DIFF
--- a/pkg/config/services.go
+++ b/pkg/config/services.go
@@ -914,8 +914,5 @@ var SupportedServices = serviceConfigs{
 		ResourceFilters: []*string{
 			aws.String("events"),
 		},
-		DimensionRegexps: []*regexp.Regexp{
-			regexp.MustCompile(":rule/(?P<EventBusName>[^/]+)/(?P<RuleName>[^/]+)$"),
-		},
 	},
 }


T0tt1:
+1

PavelDamyanov:
+1

vbabitskiy:
+1

teo9806:
It would be awesome if this is going to be resolved as early as possible.

cristiangreco (commented):
Thanks for the contribution!

Would you mind adding a test like https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/18356f659ab27c10556abae1f00f8003904770ca/pkg/job/maxdimassociator/associator_mq_test.go (or any other similar test in that folder)? It would allow to check how the regex matches resource ARNs.

HristoStoyanovYotpo:
> Thanks for the contribution!
> 
> Would you mind adding a test like https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/18356f659ab27c10556abae1f00f8003904770ca/pkg/job/maxdimassociator/associator_mq_test.go (or any other similar test in that folder)? It would allow to check how the regex matches resource ARNs.

Thank you sir! There was actually existing test that was failing, I decided to just updated the regex so that it works for both regular and partner event buses. I would expect everything to be alright now.

pkg/config/services.go
```diff
@@ -915,7 +915,7 @@ var SupportedServices = serviceConfigs{
 			aws.String("events"),
 		},
 		DimensionRegexps: []*regexp.Regexp{
-			regexp.MustCompile(":rule/(?P<EventBusName>[^/]+)/(?P<RuleName>[^/]+)$"),
+			regexp.MustCompile(":rule/(?P<EventBusName>.+)/(?P<RuleName>[^/]+)$"),
```
cristiangreco, line 918: I'm wondering whether we need a separate regex for partner events instead, like e.g.

```
		DimensionRegexps: []*regexp.Regexp{
			regexp.MustCompile(":rule/(?P<EventBusName>[^/]+)/(?P<RuleName>[^/]+)$"),
			regexp.MustCompile(":rule/aws.partner/(?P<EventBusName>.+)/(?P<RuleName>[^/]+)$"),
		},
```

Can you check if this works for you?

HristoStoyanovYotpo: Thanks @cristiangreco, according to the docs it seems like partner busses are always prepended with aws.partner: https://docs.aws.amazon.com/eventbridge/latest/onboarding/amazon_eventbridge_partner_onboarding_guide.html#AWS_EventBridge_Partner_Onboarding_Names
I'll update the PR as suggested.

cristiangreco (approved):

PR #1368, opened 2024-04-05
getmetricdata: introduce an iterator

kgeckhart:
This PR introduces an iterator to the `getmetricsdata.Processor` allowing further abstraction to how metrics batches are created. This iterator has the responsibility of creating usable batches for calling GetMetricData ensuring the `Start` and `EndTime` values are consistent for the batch. 

This abstraction helps to move towards fixing the 3 issues laid out in https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1290,
1.  Eliminating the ability to set rounding period manually and always uses the metric period
    * The current logic for rounding period can result in inconsistent `Start` and `EndTime` between batches
    * This negates its intent to ensure `Start` and `EndTime` align with the metric period for [CloudWatch best practices]
    (https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html)
image

2. This abstraction can be used to inject a batching implementation which batches by `length` + `period` combinations (have this done in a follow up) 
3. Removes the logic to add a delay at the metric level and instead logging a warning. We can choose to eliminate the field all together from the config or add support for metric delay later.

This also sets up an abstraction to build upon to accomplish https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1094 without goroutines. This will be expanded upon further in a comment in that issue.

### DIFF
--- a/pkg/job/getmetricdata/batching.go
+++ b/pkg/job/getmetricdata/batching.go
@@ -1,0 +1,97 @@
+package getmetricdata
+
+import (
+	"math"
+	"time"
+
+	"github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/model"
+)
+
+type WindowCalculator interface {
+	Calculate(period time.Duration, length time.Duration, delay time.Duration) (time.Time, time.Time)
+}
+
+type iteratorFactory struct {
+	metricsPerQuery  int
+	windowCalculator WindowCalculator
+}
+
+func (b iteratorFactory) Build(data []*model.CloudwatchData, jobMetricLength, jobMetricDelay int64) BatchIterator {
+	if len(data) == 0 {
+		return nothingToIterate{}
+	}
+
+	params := data[0].GetMetricDataProcessingParams
+	params.Length = jobMetricLength
+	params.Delay = jobMetricDelay
+
+	return NewSimpleBatchIterator(b.windowCalculator, b.metricsPerQuery, data, params)
+}
+
+type nothingToIterate struct{}
+
+func (n nothingToIterate) Size() int {
+	return 0
+}
+
+func (n nothingToIterate) Next() ([]*model.CloudwatchData, time.Time, time.Time) {
+	return nil, time.Time{}, time.Time{}
+}
+
+func (n nothingToIterate) HasMore() bool {
+	return false
+}
+
+type simpleBatchingIterator struct {
+	size            int
+	currentBatch    int
+	startTime       time.Time
+	endTime         time.Time
+	data            []*model.CloudwatchData
+	entriesPerBatch int
+}
+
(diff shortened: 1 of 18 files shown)


pkg/job/getmetricdata/batching.go
```diff
@@ -0,0 +1,97 @@
...
+	s.currentBatch++
+	return result, s.startTime, s.endTime
+}
+
+func (s *simpleBatchingIterator) HasMore() bool {
+	return s.currentBatch < s.size
+}
+
+// NewSimpleBatchIterator returns an iterator which slices the data in place based on the metricsPerQuery.
+func NewSimpleBatchIterator(windowCalculator WindowCalculator, metricsPerQuery int, data []*model.CloudwatchData, params *model.GetMetricDataProcessingParams) BatchIterator {
```
cristiangreco, line 82: Wondering if we can remove the dependency on `WindowCalculator`, e.g. move it to the processor, calculate `startTime` and `endTime` in the processor and pass those two fixed values to the iterator.

kgeckhart: I could move the calculator out to the processor but I would still want to have the iterator return `([]*model.CloudWatchData, *model.GetMetricDataProcessingParams)` so its clearly provides the params for the batch. Otherwise, the processor has to index the first element for the params to get them for start + endtime and that is kinda awkward. 

I think it's important iterator owns the params like this though and it's the basis of the follow up which can produce more "correct" batches https://github.com/kgeckhart/yet-another-cloudwatch-exporter/compare/kgeckhart/getmetricdata-iterator...kgeckhart/batching-variable-start-end-time?expand=1

cristiangreco: IIUC though the core of the logic is in the factory.Build. Even the varyingTimeIterator doesn't use `windowCalculator` or `startTime` and `endTime` after they're set at creation time.

I guess what I see here is that iterator doesn't care much of start and end, that's why I was thinking that bit could be isolated more.

kgeckhart: I agree, I think the start + endTime calc can be moved to `processor`. I'm only highlighting the `iterator` should be responsible for saying "here's the params which need to be used" so instead of,

```
Next() ([]*model.CloudwatchData, time.Time, time.Time)
Next() ([]*model.CloudwatchData)
```
we get

```
//  "here's the next batch and the params for it"
Next() ([]*model.CloudwatchData, *model.GetMetricDataProcessingParams)
```

kgeckhart:
Replaced by https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1389 and https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1388

PR #1376, opened 2024-04-08
fix association with gwlb

vainiusd:
Bug defined in https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1345


pkg/config/services.go
```diff
@@ -444,8 +444,7 @@ var SupportedServices = serviceConfigs{
 		Namespace: "AWS/GatewayELB",
 		Alias:     "gwlb",
 		ResourceFilters: []*string{
-			aws.String("elasticloadbalancing:loadbalancer/gwy"),
-			aws.String("elasticloadbalancing:targetgroup"),
+			aws.String("elasticloadbalancing:loadbalancer"),
```
cristiangreco, line 447: Why is this removing the targetgroup regex as well?

PR #1379, opened 2024-04-10
promutil: rewrite sanitisation funcs for memory optimisation

cristiangreco:
Rewrite `sanitize` to attempt to modify the string in place, if possible.
Rewrite `splitString` to avoid using a regex.

```
% benchstat before.txt after.txt
goos: darwin
goarch: arm64
pkg: github.com/nerdswords/yet-another-cloudwatch-exporter/pkg/promutil
                 │ before.txt  │              after.txt              │
                 │   sec/op    │   sec/op     vs base                │
_BuildMetrics-12   201.1µ ± 2%   114.5µ ± 6%  -43.07% (p=0.000 n=10)

                 │  before.txt  │              after.txt               │
                 │     B/op     │     B/op      vs base                │
_BuildMetrics-12   5.750Ki ± 1%   4.281Ki ± 0%  -25.55% (p=0.000 n=10)

                 │ before.txt  │             after.txt              │
                 │  allocs/op  │ allocs/op   vs base                │
_BuildMetrics-12   168.00 ± 0%   95.00 ± 0%  -43.45% (p=0.000 n=10)
```

### DIFF
--- a/pkg/promutil/prometheus.go
+++ b/pkg/promutil/prometheus.go
@@ -3,8 +3,8 @@ package promutil
 import (
 	"strings"
 	"time"
+	"unsafe"
 
-	"github.com/grafana/regexp"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 )
@@ -90,7 +90,6 @@ var replacer = strings.NewReplacer(
 	">", "_",
 	"%", "_percent",
 )
-var splitRegexp = regexp.MustCompile(`([a-z0-9])([A-Z])`)
 
 type PrometheusMetric struct {
 	Name             *string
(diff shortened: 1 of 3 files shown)


pkg/promutil/prometheus.go
```diff
@@ -155,10 +154,54 @@ func PromStringTag(text string, labelsSnakeCase bool) (bool, string) {
...
+	}
+
+	b := []byte(text)
+	for i := 0; i < len(b); i++ {
+		switch b[i] {
+		case ' ', ',', '\t', '/', '\\', '.', '-', ':', '=', '@', '<', '>':
+			b[i] = '_'
+		}
+	}
+	return *(*string)(unsafe.Pointer(&b))
```
thepalbi, line 173: this is more efficient than doing `string(b)`?

thepalbi (approved):
🔥

pkg/promutil/prometheus.go
cristiangreco: Good point, I don't think unsafe is needed anymore, I was playing with it in a previous iteration. Shouldn't change much in terms of performance, but `string(b)` is easier to read.

PR #1387, opened 2024-04-18
Do not allow using aliases as job types in discovery jobs

cristiangreco:
This is a breaking change in the config parsing that doesn't allow to use namespace aliases anymore. Full AWS namespaces should be used instead. The error message should be enough to understand how to fix the config file.

This change should bring consistency in the name of metrics being generated (when using the alias, the metric prefix might change compared to when using the namespace). This standardisation will also open the door to better docs and perf improvements.

Relates to https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1131

### DIFF
--- a/examples/qldb.yml
+++ b/examples/qldb.yml
@@ -1,10 +1,10 @@
 apiVersion: v1alpha1
 discovery:
   exportedTagsOnMetrics:
-    qldb:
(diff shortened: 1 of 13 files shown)


kgeckhart (approved):
👏

PR #1405, opened 2024-05-06
feat: add AWS/RUM as supported service

hexionas:
Added this locally and ran using docker-compose. Correctly fetches the metric as seen below.

```yaml
    - type: AWS/RUM
      regions: [eu-west-1]
      period: 60
      length: 60
      metrics:
        - name: JsErrorCount
          statistics:
            - Sum
```

```
aws_rum_js_error_count_sum{account_id="<redacted>", dimension_application_name="portal-monitor-dev", instance="yace:8080", job="yace", name="global", region="eu-west-1"}
```


hexionas:
One thing i've noticed. If I don't set `nilToZero: true` on the metric coming from RUM then it stays at `NaN` even if a metric is present.

PR #1412, opened 2024-05-13
Add support for batching by time params

kgeckhart:
This PR introduces the ability to batch the `GetMetricData` requests by the params which directly influence `StartTime` and `EndTime` vs allowing them to vary per request batch. This is a backwards incompatible change which removes the usage of the config rounding period.

Resolves https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1290


pkg/model/model.go
```diff
@@ -54,16 +53,6 @@ type CustomNamespaceJob struct {
...
-	JobLevelMetricFields
-}
-
-type JobLevelMetricFields struct {
-	Statistics             []string
-	Period                 int64
-	Length                 int64
-	Delay                  int64
-	NilToZero              *bool
-	AddCloudwatchTimestamp *bool
```
kgeckhart, line 66: All of these fields are pushed down to the metric level and were no longer referenced

cristiangreco (approved):
Great work!!!

PR #1430, opened 2024-05-31
Add uid to the dashboards based on md5 of the filenames

andriikushch:
This PR adds UID to the dashboards for the AWS services.

### DIFF
--- a/mixin/dashboards/all.libsonnet
+++ b/mixin/dashboards/all.libsonnet
@@ -1,7 +1,17 @@
 {
-  'ebs.json': import 'ebs.libsonnet',
-  'ec2.json': import 'ec2.libsonnet',
-  'lambda.json': import 'lambda.libsonnet',
-  'rds.json': import 'rds.libsonnet',
-  's3.json': import 's3.libsonnet',
+  'ebs.json': (import 'ebs.libsonnet') {
+    uid: std.md5('ebs.libsonnet'),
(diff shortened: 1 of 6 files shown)


cristiangreco (approved):

mixin/dashboards/all.libsonnet
```diff
@@ -1,7 +1,17 @@
...
+    uid: std.md5('ec2.libsonnet'),
+  },
+  'lambda.json': (import 'lambda.libsonnet') {
+    uid: std.md5('lambda.libsonnet'),
+  },
+  'rds.json': (import 'rds.libsonnet') {
+    uid: std.md5('rds.libsonnet'),
+  },
+  's3.json': (import 's3.libsonnet') {
+    uid: std.md5('s3.libsonnet'),
```
thepalbi, line 15: Mmm so I think what we needed is not a movable md5 hash as uid, since that might change in the future, but a fixed one, much like [here for example](https://github.com/grafana/cloud-onboarding/blob/dfa91b9846441b83c482ce29f03c8315f6d898f4/integrations/gcp-logs/dashboards.libsonnet#L7),

which allows a dashboard URL to be something like 

```
https://test.grafana.net/d/gcp-logs-overview/gcp-logs-overview?orgId=1
```

which is fixed, despite changes on the dashboard

PR #1432, opened 2024-06-03
Start a unified scraper

kgeckhart:
This PR starts a new unfinished code path for scraping discovery and custom namespaces jobs which is intended to,

* Unify the code which scrape both job types
* Propagate errors from individual job runs out of the scrapper to be handled by the caller
* Be unit testable (this is 50% of the PR)

This PR starts introducing pieces of the implementation one at a time in digestible chunks with proper tests. The meaty changes are the entry point for the new scraper in`scraper.go` which will be injected into the existing [scrape.go](https://github.com/nerdswords/yet-another-cloudwatch-exporter/compare/master...kgeckhart:yet-another-cloudwatch-exporter:kgeckhart/new-job-runner?expand=1#diff-917daa317df5c7ef2802bd99d95f06921e3ca7ec53170f93f9226e2c6447d1ac) (it's the full no-test implementation if you're interested in skipping ahead). 

Related to https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/1094
There's another issue I need to open which this is a piece of related to splitting CloudWatch and ResourceMetadata jobs in to separate things.


pkg/job/scraper.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (e Error) ToLoggerKeyVals() []interface{} {
+	return []interface{}{
+		"account_id", e.AccountID,
+		"namespace", e.Namespace,
+		"region", e.Region,
+		"role_arn", e.RoleARN,
+	}
+}
```
kgeckhart, line 240: Rather customer error type to ensure we don't miss contextual logging values we previously had when error logging happened during scraping.

pkg/job/scraper_test.go
```diff
@@ -0,0 +1,454 @@
...
+			changelog, err = diff.Diff(tc.expectedMetrics, metrics)
+			assert.NoError(t, err, "failed to diff metrics")
+			assert.Len(t, changelog, 0, changelog)
+
+			// We don't need the check the exact error or message
+			changelog, err = diff.Diff(tc.expectedErrs, errs, diff.Filter(func(_ []string, _ reflect.Type, field reflect.StructField) bool {
+				return !(field.Name == "Err" || field.Name == "Message")
+			}))
+			assert.NoError(t, err, "failed to diff errs")
+			assert.Len(t, changelog, 0, changelog)
```
kgeckhart, line 451: This uses a struct diff library because of the complexity of the struct being compared. The library can diff two complex structs and by default will ignore slice order on any fields which are a slice type which drastically simplifies the assertions here.

pkg/job/scraper.go
```diff
@@ -0,0 +1,240 @@
...
+	resourceResults := make([]model.TaggedResourceResult, 0)
+	s.logger.Debug("Starting job runs")
+
+	jobConfigVisitor(s.jobsCfg, func(job any, role model.Role, region string) {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			var namespace string
+			jobAction(s.logger, job, func(job model.DiscoveryJob) {
```
tristanburgess, line 76: I think that `jobAction()` should be defined above this `Scrape()` method, since it's very small, and helps make it more clear that it's just running one of these inline functions based on the job type

pkg/job/scraper.go
```diff
@@ -0,0 +1,240 @@
...
+			jobLogger := s.logger.With("namespace", namespace, "region", region, "arn", role.RoleArn)
+
+			accountID := roleRegionToAccount[role][region]()
+			if accountID == "" {
+				jobError := Error{
+					AccountID: "",
+					Region:    region,
+					RoleARN:   role.RoleArn,
+					Namespace: namespace,
+					Message:   "Account for job was not found see previous errors",
```
tristanburgess, line 90:
```suggestion
					Message:   "Account for job was not found. There should be a corresponding 'Failed to get Account' error logged previously.",
```

Just to hopefully tie it a bit more specifically so that operators will know what they're looking for.

tristanburgess: Alternatively perhaps the `roleRegionToAccount` mapped function could have a return signature of `(string, err)` and directly check for `if err != nil` and add the error in this struct?

pkg/job/scraper.go
```diff
@@ -0,0 +1,240 @@
...
+				jobLogger.Debug("Ending job run early due to job error see job errors")
+				return
+			}
+			jobLogger.Debug("Starting cloudwatch metrics runner")
+			cwRunner := s.runnerFactory.NewCloudWatchRunner(jobLogger, region, role, jobToRun)
+			metricResult, err := cwRunner.Run(ctx)
+			if err != nil {
+				jobError := Error{
+					AccountID: accountID,
+					Region:    region,
```
tristanburgess, line 152: I noticed that a few of the params are reused in this struct throughout this function, maybe a single error instance should be built at the beginning, add the AccountID when it exists, and modify the `Message` and `Err` params as needed?

kgeckhart: Since the error isn't necessary for most job runs I'm fine with leaving creation for when we need it vs creating one and throwing it away most of the time.

pkg/job/scraper.go
kgeckhart: I like it!

pkg/job/scraper.go
kgeckhart: I ended up taking this a different direction where instead of building the Error type I build up a context object. This object is then morphed to what's needed in the return hiding a lot of the distracting mapping logic.

tristanburgess (approved):
🚀 LGTM! Definitely might want another approval though.

pkg/job/scraper.go
```diff
@@ -0,0 +1,241 @@
...
+					} else {
+						jobLogger.Debug("No tagged resources")
+					}
+					jobLogger.Debug("Resource discovery finished", "number_of_discovered_resources", len(resources))
+
+					jobToRun = cloudwatchrunner.DiscoveryJob{Job: job, Resources: resources}
+				}, func(job model.CustomNamespaceJob) {
+					jobToRun = cloudwatchrunner.CustomNamespaceJob{Job: job}
+				},
+			)
```
kgeckhart, line 129: I wanted to keep resource discovery here and somewhat obvious. In the future when resource discovery happens elsewhere, I would expect this to be something like

```suggestion
			jobAction(jobLogger, job,
				func(job model.DiscoveryJob) {
					jobToRun = cloudwatchrunner.DiscoveryJob{Job: job}
				}, func(job model.CustomNamespaceJob) {
					jobToRun = cloudwatchrunner.CustomNamespaceJob{Job: job}
				},
			)
```
I might need to pass through a new `ResourceProvider` type but ideally can mask it through the ResourceEnrichment pipeline that will be setup.

pkg/job/scraper.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (s Scraper) Scrape(ctx context.Context) ([]model.TaggedResourceResult, []model.CloudwatchMetricResult, []Error) {
+	// Setup so we only do one GetAccount call per region + role combo when running jobs
+	roleRegionToAccount := map[model.Role]map[string]func() (string, error){}
+	jobConfigVisitor(s.jobsCfg, func(_ any, role model.Role, region string) {
+		if _, exists := roleRegionToAccount[role]; !exists {
+			roleRegionToAccount[role] = map[string]func() (string, error){}
+		}
+		roleRegionToAccount[role][region] = sync.OnceValues[string, error](func() (string, error) {
```
cristiangreco, line 53: (nit) I think it might happen that a func for a given role+region has already been assigned as part of the sequential visitor work, we could check if the entry already exists?

pkg/job/scraper.go
```diff
@@ -0,0 +1,241 @@
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			var namespace string
+			jobAction(s.logger, job, func(job model.DiscoveryJob) {
+				namespace = job.Type
+			}, func(job model.CustomNamespaceJob) {
+				namespace = job.Namespace
+			})
```
cristiangreco, line 79: Wondering if it's time to go and rename `job.Type` instead of doing this? (it's an internal model after all).

cristiangreco: More generally, maybe instead of `jobAction` we can use a minimal interface through the models you created in `cloudwatchrunner.DiscoveryJob` and `cloudwatchrunner.CustomNamespaceJob`?

pkg/job/scraper_test.go
```diff
@@ -0,0 +1,454 @@
+package job_test
```
cristiangreco, line 1: Great work!

pkg/job/scraper_test.go
```diff
@@ -0,0 +1,454 @@
...
+							MetricName:          "metric-2",
+							ResourceName:        "resource-2",
+							Namespace:           "custom-namespace",
+							Dimensions:          []model.Dimension{{Name: "dimension2", Value: "value2"}},
+							GetMetricDataResult: &model.GetMetricDataResult{Statistic: "Minimum", Datapoint: aws.Float64(2.0), Timestamp: time.Time{}},
+						},
+					},
+				},
+			},
+			expectedErrs: []job.Error{
```
cristiangreco, line 350: Could we match on the error message as well in these error cases tests?

cristiangreco: Ok I just read the comment later 

> We don't need the check the exact error or message

Why is that the case?

pkg/job/scraper_test.go
```diff
@@ -0,0 +1,454 @@
...
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			rf := testRunnerFactory{
+				GetAccountFunc:    tc.getAccountFunc,
+				MetadataRunFunc:   tc.metadataRunFunc,
+				CloudwatchRunFunc: tc.cloudwatchRunFunc,
+			}
+			sr := job.NewScraper(logging.NewLogger("", true), tc.jobsCfg, &rf)
```
cristiangreco, line 435: (nit) might use the noop logger here.

cristiangreco (commented):
Great stuff, left a bunch of comments for discussion.

pkg/job/scraper_test.go
kgeckhart: I tend to avoid checking specific error messages unless it's pivital that an error show up with certain text it tends to make tests super brittle. If there's errors you think are really valuable I can type them and would feel better about checking types vs just text.

pkg/job/scraper_test.go
kgeckhart: Funny enough, I did this because it was helpful for debugging tests without needing to go to the level of attaching a debugger. I can switch it though it it's too much noise in the test logs.

pkg/job/scraper_test.go
cristiangreco: Yeah checking types is a good option. I think that while at it, we could use this opportunity to start being more meticolous around errors that bubble up.

pkg/job/scraper_test.go
cristiangreco: Nope, not a big deal. 👍

pkg/job/scraper_test.go
kgeckhart: I added this, https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/1432/files#diff-bf5be17ad3e7eb74bb4839c9cd988e47b9f3bc04a4debae6b12d3e0e24b3dad5R46-R52, which would be a way for us to give some of the more important use cases a type to match against. WDYT?

cristiangreco (approved):

PR #1538, opened 2024-10-08
Fix error metrics for APIGateway example

vasyharan:
Updates the APIGateway example with the correct names for error metrics

### DIFF
--- a/examples/apigwv2.yml
+++ b/examples/apigwv2.yml
@@ -1,0 +1,18 @@
+apiVersion: v1alpha1
+discovery:
+  jobs:
+    - type: AWS/ApiGateway
+      regions:
+        - us-east-1
+      period: 300
+      length: 300
+      metrics:
+        - name: Latency
+          statistics: [Average, Maximum, p95, p99]
+        - name: Count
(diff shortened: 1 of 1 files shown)


examples/apigw.yml
```diff
@@ -11,7 +11,7 @@ discovery:
           statistics: [Average, Maximum, p95, p99]
         - name: Count
           statistics: [SampleCount, Sum]
-        - name: 4xx
+        - name: 4XXError
           statistics: [Sum]
-        - name: 5xx
+        - name: 5XXError
```
cristiangreco, line 16: @vasyharan I think `4xx` and `5xx` are valid metrics for APIGateway v1 (http), while `4XXError` and `5XXError` are for APIGateway v2 (rest). Maybe we should have two example files?

cristiangreco (changes requested):
Left a comment for clarification.

examples/apigw.yml
vasyharan: I didn't realize that these were different versions! Updated to add a new files for v2.

PR #1577, opened 2024-11-20
Update Logger interface

nick-klisch:
To optionally reduce the level of info-level logging

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -57,6 +57,7 @@ var (
 	addr                  string
 	configFile            string
 	debug                 bool
+	reduceInfoLogs        bool
 	logFormat             string
 	fips                  bool
 	cloudwatchConcurrency cloudwatch.ConcurrencyConfig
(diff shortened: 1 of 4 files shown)


SuperQ:
FYI, we're going to migrate from go-kit/log to the Go standard slog library soon.

rhysxevans:
Hi, is there a timeline on this ? and would it inplement standard nomenclature like `log-level` not requiring things like `IsDebugEnabled`. IE I think, personally, we should be abel to do soemthing like `--log-level=debug` and we should get all debug and above and `--log-level=info` should only show info and above (ie not debug) . 

And I use the word standard loosely as I am not sure what the go standard is 

Relates to: https://github.com/prometheus-community/yet-another-cloudwatch-exporter/discussions/1573

SuperQ:
That's the default way we do things. This exporter seems to have some extra logging code that can probably be simplified.

CC @tjhop

tjhop:
PR has been opened to convert to slog, please feel free to take a look. Feedback welcome.

PR #1586, opened 2024-12-03
ref!: adopt log/slog + other prom configs, drop custom logging pkg

tjhop:
I would call this a breaking change, as there are updates to command line flag interactions.

Changes included:
- convert `--debug` bool flag -> `--log.level` string flag. this aligns with prom conventions. This project uses urfave/cli for application commands/flags rather than kingpin as other go packages do, so I couldn't use the promslog/flag pkg directly. Rather, I use our upstream promslog/kingpin configs in the configs we pass here for urfave/cli. We can decide to move off urfave/cli and/or adopt kingpin in the future.
- removes `pkg/logging` in favor of using `prometheus/common/promslog`. A small utility function is included in main to simplify some boilerplate with logger initialization, since we instantiate in main in a few places. All functions have been converted to accept/use *slog.Loggers rather than the custom logger interface.
- the removal of the custom logging pkg also removes the `IsDebugEnabled()` method, but this is fine because the logger is leveled; debug logs will not print unless `--log.level debug` is passed at the command line, there is no need to check if debug is enabled at log time in the vast majority of cases.
- Updates tests as needed; updates nop loggers, test signatures that used `IsDebugEnabled()`, etc.
- enable sloglint in golangci-lint, remove go-kit/log configs


tjhop:
Supercedes #1577 

cc: @SuperQ

SuperQ:
Thanks, we'll probably slowly convert over other things like `urfave/cli` to Prometheus community normal components over time.

SuperQ (approved):
Great work!

tjhop:
Looking into the config verification failure in tests

tjhop:
Alright, I fixed up the verification test; panic was caused by a typo/wrong variable reference. Sorted out now :+1:

tjhop:
Bump, lemme know if there's any other feedback :+1:

SuperQ:
Thanks!

PR #1607, opened 2025-01-03
CI workflow: save_artifacts after build

cristiangreco:
Seems that publish_main expects to find build artifacts.


SuperQ:
Thanks!

Issue #1615, opened 2025-01-06
[BUG] Version is not properly embedded into go binary


asherf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### YACE version

0.62.1

### Current Behavior

log lines on startup:

{"time":"2025-01-06T17:09:52.976480206Z","level":"INFO","source":"main.go:259","msg":"Parsing config","version":"custom-build"}
{"time":"2025-01-06T17:09:52.977881278Z","level":"INFO","source":"main.go:344","msg":"Yace startup completed","version":"custom-build","version":"custom-build","feature_flags":"aws-sdk-v2"}

### Expected Behavior

{"time":"2025-01-06T17:09:52.976480206Z","level":"INFO","source":"main.go:259","msg":"Parsing config","version":"0.62.1"}
{"time":"2025-01-06T17:09:52.977881278Z","level":"INFO","source":"main.go:344","msg":"Yace startup completed","version":"0.62.1","version":"custom-build","feature_flags":"aws-sdk-v2"}

### Anything else?

this worked before the transition of the repo to the new GH org.
so this is likely a side effect of this transition.

PR #1616, opened 2025-01-06
Use Prometheus common version library

SuperQ:
Use the Prometheus common version library to provide embedded build context.

Fixes: https://github.com/prometheus-community/yet-another-cloudwatch-exporter/issues/1615

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/prometheus/common/promslog"
 	promslogflag "github.com/prometheus/common/promslog/flag"
+	"github.com/prometheus/common/version"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/sync/semaphore"
 
@@ -47,8 +48,6 @@ Version: %s
 </html>`
 	htmlPprof = `<p><a href="/debug/pprof">Pprof</a><p>`
 )
-
-var version = "custom-build"
 
 var sem = semaphore.NewWeighted(1)
 
@@ -90,7 +89,7 @@ func main() {
 func NewYACEApp() *cli.App {
 	yace := cli.NewApp()
 	yace.Name = "Yet Another CloudWatch Exporter"
-	yace.Version = version
+	yace.Version = version.Version
 	yace.Usage = "YACE configured to retrieve CloudWatch metrics through the AWS API"
 	yace.Description = ""
 	yace.Authors = []*cli.Author{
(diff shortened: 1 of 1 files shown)


cmd/yace/main.go
```diff
@@ -90,7 +89,7 @@ func main() {
 func NewYACEApp() *cli.App {
 	yace := cli.NewApp()
 	yace.Name = "Yet Another CloudWatch Exporter"
-	yace.Version = version
+	yace.Version = version.Version
```
cristiangreco, line 92: IIUC promu will read the version from the `VERSION` file? This is really handy at release time, but is misleading when building an unreleased version. Previously we had the makefile setup so that it would output the current branch and revision, something like this:

```makefile
GIT_BRANCH   ?= $(shell git rev-parse --abbrev-ref HEAD)
GIT_REVISION ?= $(shell git rev-parse --short HEAD)
VERSION      ?= $(GIT_BRANCH)-$(GIT_REVISION)
```

How could we reproduce a similar behaviour?

cmd/yace/main.go
```diff
@@ -341,7 +340,7 @@ func startScraper(c *cli.Context) error {
 		go s.decoupled(ctx, logger, newJobsCfg, cache)
 	})
 
-	logger.Info("Yace startup completed", "version", version, "feature_flags", strings.Join(featureFlags, ","))
+	logger.Info("Yace startup completed", "build_context", version.BuildContext(), "feature_flags", strings.Join(featureFlags, ","))
```
cristiangreco, line 343: Just double-checking, did you intend to put the whole buildcontext as a single value? The output looks like:

```
"build_context":"(go=go1.23.4, platform=darwin/arm64, user=cristian@MacG2.local, date=20250107-08:53:15, tags=unknown)"
```

cmd/yace/main.go
SuperQ: I think this would be a feature request for promu. The version package vars are injected by the `.promu.yml` file.

cmd/yace/main.go
SuperQ: Yes, that's typically how we've done it in other exporters.

For example:
```
time=2025-01-07T13:54:03.868+01:00 level=INFO source=/home/ben/src/prometheus-community/bind_exporter/bind_exporter.go:559 msg="Starting bind_exporter" version="(version=0.7.0, branch=superq/slog, revision=ea563db0a26ecc5c32bd36f0770ca81ef2c8be9c)"
```

SuperQ:
IMO, the version string itself is not all that important compared to the build context. This is why I added the build context output to the log.

I was also going to add a a build info metric with the [version collector package](https://pkg.go.dev/github.com/prometheus/client_golang@v1.20.5/prometheus/collectors/version). But I hadn't decided on where exactly I wanted to register it.

cmd/yace/main.go
cristiangreco: Maybe we could add `version.Info()` in this very log line next to the build context, so at least the branch and revision are outputted somehow.

SuperQ:
```json
{"time":"2025-01-08T17:36:17.014336719Z","level":"INFO","source":"main.go:343","msg":"Yace startup completed","version":"0.62.1","version":"0.62.1","build_context":"(go=go1.23.3, platform=linux/amd64, user=ben@mrhat, date=20250108-17:36:06, tags=unknown)","feature_flags":""}
```

PR #1621, opened 2025-01-14
Add example of Lambda Edge

tyagian:
Since we can't use Lambda@Edge as a discovery job, we can add it as static job. 
Not everyone may know about using as static job and it can be confusing when example configuration for Lambda don't work. 

This can help them.


cristiangreco (approved):

cristiangreco:
@tyagian thanks! Could you please sign the commit?

tyagian:
RCO check was failing because old commit was missing sign so I created new PR
https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1628

PR #1642, opened 2025-02-25
feat: reimplement history data export

woehrl01:


pkg/config/config.go
```diff
@@ -387,6 +389,19 @@ func (m *Metric) validateMetric(logger *slog.Logger, metricIdx int, parent strin
...
+	mAddHistoricalMetrics := m.AddHistoricalMetrics
+	if mAddHistoricalMetrics == nil {
+		if discovery != nil && discovery.AddHistoricalMetrics != nil {
+			mAddHistoricalMetrics = discovery.AddHistoricalMetrics
+		} else {
+			mAddHistoricalMetrics = aws.Bool(false)
+		}
+	}
+
+	if aws.BoolValue(mAddHistoricalMetrics) && !aws.BoolValue(mAddCloudwatchTimestamp) {
```
jotweh, line 401: it would be great to have this dependency and the reason for it documented in configuration.md. As I understand it, the reason is that prometheus would otherwise potentially ignore historic data points because it would set the timestamp when retrieving data from yace?

jotweh:
Could you also adjust one of the examples with the new `AddHistoricalMetrics` flag and different values for `period` and `length`?

pkg/config/config.go
woehrl01: added the information. yes your right, if you don't set cloudwatch timestamp the data would be deduplicated during export and thus the last value would be shown. which would lead to more confusion, so better be explicit on both things enabled.

docs/configuration.md
```diff
@@ -137,6 +137,13 @@ statistics:
 # Export the metric with the original CloudWatch timestamp (General Setting for all metrics in this job)
 [ addCloudwatchTimestamp: <boolean> ]
 
+# Enables the inclusion of past metric data points from the CloudWatch response if available.
+# This is useful when a metric is configured with a 60-second period and a 300-second duration, ensuring that all
+# five data points are exposed at the metrics endpoint instead of only the latest one.
+# Note: This option requires `addCloudwatchTimestamp` to be enabled
+# (General Setting for all metrics in this job)
+[ addHistoricalMetrics: <boolean> ]
```
kgeckhart, line 145: WDYT about calling it `exportAllDataPoints`? I worry with a name like `addHistoricalMetrics` it the feature could be confused for a historical data extract.

docs/configuration.md
```diff
@@ -137,6 +137,13 @@ statistics:
 # Export the metric with the original CloudWatch timestamp (General Setting for all metrics in this job)
 [ addCloudwatchTimestamp: <boolean> ]
 
+# Enables the inclusion of past metric data points from the CloudWatch response if available.
+# This is useful when a metric is configured with a 60-second period and a 300-second duration, ensuring that all
+# five data points are exposed at the metrics endpoint instead of only the latest one.
+# Note: This option requires `addCloudwatchTimestamp` to be enabled
```
kgeckhart, line 143: I think it will also require that the metrics destination you are using has support for out of order timestamps. Probably worth a call out to at least the prometheus docs about it https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tsdb.

pkg/promutil/migrate.go
```diff
@@ -101,40 +101,53 @@ func BuildMetrics(results []model.CloudwatchMetricResult, labelsSnakeCase bool,
...
+					output = append(output, &PrometheusMetric{
+						Name:             name,
+						Labels:           promLabels,
+						Value:            exportedDatapoint,
+						Timestamp:        ts,
+						IncludeTimestamp: metric.MetricMigrationParams.AddCloudwatchTimestamp,
+					})
+
+					if !metric.MetricMigrationParams.AddHistoricalMetrics {
+						// If we're not adding historical metrics, we can skip the rest of the statistics for this metric
```
kgeckhart, line 147:
```suggestion
						// If we're not adding historical metrics, we can skip the rest of the datapoints for this metric
```

pkg/promutil/migrate.go
```diff
@@ -101,40 +101,53 @@ func BuildMetrics(results []model.CloudwatchMetricResult, labelsSnakeCase bool,
...
+					if err != nil {
+						return nil, nil, err
+					}
+					var exportedDatapoint float64
+					if dataPoint == nil && metric.MetricMigrationParams.AddCloudwatchTimestamp {
+						// If we did not get a datapoint then the timestamp is a default value making it unusable in the
+						// exported metric. Attempting to put a fake timestamp on the metric will likely conflict with
+						// future CloudWatch timestamps which are always in the past. It's safer to skip here than guess
+						continue
+					}
```
kgeckhart, line 117: I think there's a slight change in behavior here, if the feature is disabled and the first datapoint is nil we will bypass the break at the end potentially exporting an old data point as current. Could you add a test for this use case to see if it happens?

pkg/clients/cloudwatch/v2/client.go
```diff
@@ -149,10 +149,15 @@ func (c client) GetMetricData(ctx context.Context, getMetricData []*model.Cloudw
...
-			mappedResult.Timestamp = metricDataResult.Timestamps[0]
+		mappedResult := cloudwatch_client.MetricDataResult{
+			ID:         *metricDataResult.Id,
+			Datapoints: make([]cloudwatch_client.DatapointWithTimestamp, 0, len(metricDataResult.Timestamps)),
+		}
+		for i := 0; i < len(metricDataResult.Timestamps); i++ {
+			mappedResult.Datapoints = append(mappedResult.Datapoints, cloudwatch_client.DatapointWithTimestamp{
+				Datapoint: &metricDataResult.Values[i],
+				Timestamp: metricDataResult.Timestamps[i],
+			})
```
kgeckhart, line 160: WDYT about using the NewDataPoint function here?

pkg/clients/cloudwatch/v1/client.go
```diff
@@ -143,10 +143,14 @@ func (c client) GetMetricData(ctx context.Context, getMetricData []*model.Cloudw
...
-			mappedResult.Datapoint = metricDataResult.Values[0]
-			mappedResult.Timestamp = *metricDataResult.Timestamps[0]
+		mappedResult := cloudwatch_client.MetricDataResult{
+			ID:         *metricDataResult.Id,
+			Datapoints: make([]cloudwatch_client.DatapointWithTimestamp, 0, len(metricDataResult.Timestamps))}
+		for i := 0; i < len(metricDataResult.Timestamps); i++ {
+			mappedResult.Datapoints = append(mappedResult.Datapoints, cloudwatch_client.DatapointWithTimestamp{
+				Datapoint: metricDataResult.Values[i],
+				Timestamp: *metricDataResult.Timestamps[i],
+			})
```
kgeckhart, line 153: Same comment here about using `NewDataPoint`

kgeckhart (commented):
Great work on updating this! I had a few minor comments and will see about getting some more attention on it.

pkg/promutil/migrate.go
woehrl01: Your right, there must be a different behaviour here depending if we want to export historic data. Thanks for pointing that out!

woehrl01:
@kgeckhart anything I can help with?

pkg/config/config.go
```diff
@@ -154,7 +156,7 @@ func (c *ScrapeConf) Load(file string, logger *slog.Logger) (model.JobsConfig, e
 
 func (c *ScrapeConf) Validate(logger *slog.Logger) (model.JobsConfig, error) {
 	if c.Discovery.Jobs == nil && c.Static == nil && c.CustomNamespace == nil {
-		return model.JobsConfig{}, fmt.Errorf("At least 1 Discovery job, 1 Static or one CustomNamespace must be defined")
+		return model.JobsConfig{}, fmt.Errorf("at least 1 Discovery job, 1 Static or one CustomNamespace must be defined")
```
cristiangreco, line 159: Nitpick: this change is unrelated to the PR.

pkg/clients/cloudwatch/v2/client.go
```diff
@@ -149,10 +149,12 @@ func (c client) GetMetricData(ctx context.Context, getMetricData []*model.Cloudw
...
 	for _, metricDataResult := range resp.MetricDataResults {
-		mappedResult := cloudwatch_client.MetricDataResult{ID: *metricDataResult.Id}
-		if len(metricDataResult.Values) > 0 {
-			mappedResult.Datapoint = &metricDataResult.Values[0]
-			mappedResult.Timestamp = metricDataResult.Timestamps[0]
+		mappedResult := cloudwatch_client.MetricDataResult{
+			ID:         *metricDataResult.Id,
+			Datapoints: make([]cloudwatch_client.DatapointWithTimestamp, 0, len(metricDataResult.Timestamps)),
+		}
+		for i := 0; i < len(metricDataResult.Timestamps); i++ {
```
cristiangreco, line 156: I feel this new behaviour should be used only if `exportAllDataPoints` is enabled. If not, we should keep the previous behaviour and avoid copying around the additional values/timestamps from the CW response.

cristiangreco (changes requested):

cristiangreco (changes requested):
Hey, thanks for the PR! I've left a few comments, please let me know what you think. I'd like to have another pass after those points are addressed.

pkg/clients/cloudwatch/client.go
```diff
@@ -55,12 +55,26 @@ type ConcurrencyLimiter interface {
...
+func NewDataPoint(datapoint *float64, timestamp time.Time) DatapointWithTimestamp {
+	return DatapointWithTimestamp{
+		Timestamp: timestamp,
+		Datapoint: datapoint,
+	}
+}
+
+func SingleDataPoint(datapoint *float64, timestamp time.Time) []DatapointWithTimestamp {
+	return []DatapointWithTimestamp{NewDataPoint(datapoint, timestamp)}
+}
```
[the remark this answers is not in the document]
woehrl01: I don't mind, I believe this improves readability especially in the tests.

pkg/clients/cloudwatch/client.go
```diff
@@ -55,12 +55,26 @@ type ConcurrencyLimiter interface {
 }
 
 type MetricDataResult struct {
-	ID string
-	// A nil datapoint is a marker for no datapoint being found
+	ID         string
+	Datapoints []DatapointWithTimestamp
+}
+
+type DatapointWithTimestamp struct {
```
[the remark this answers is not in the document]
woehrl01: It's a data point with a timestamp, there is already a data point without a timestamp. I'm happy for any suggestion of a better name. But I think this is the most expressive one and easy to understand. What do you suggest?

pkg/clients/cloudwatch/v2/client.go
cristiangreco: I think Kyle's comment was mainly around allowing the client to return a slice of Datapoint + Timestamp, but he can confirm.

My concern is around useless mem allocations when `exportAllDataPoints` is not enabled. I think the option can be made available to the aws client, e.g. via `CloudwatchData` or `GetMetricDataProcessingParams`.

pkg/config/config.go
cristiangreco: What lint error is this solving?

pkg/clients/cloudwatch/v2/client.go
woehrl01: I'm pretty sure there is no significant useless memory allocation. All the values are already in memory because it's part of the api response. So we just allocate the slice a bit bigger, because we use a struct of slice and not a reference to the struct slice, there is a single memory alloc for the underlying array, everything else is just copying over values. So this just keeps the already returned data for a few function calls longer in memory.

Feels like an unessary add of complexity, to pass around those parameters and adding additional tests for validate that behaviour.

woehrl01: Following up on my earlier comment: even if we assume returning a slice of 5 structs increases memory usage, the overhead is still reasonable, around ~128 MB for 1 million metrics. That comes from storing 4 additional data points per metric, each at 32 bytes (a time.Time value + a *float64 pointer). It's just a larger backing array and a few more pointers to scan during GC. This only becomes relevant if we're allocating and retaining millions of these slices, which we're not.

And if we were dealing with that many, querying 1 million CloudWatch metrics, the real issue wouldn't be memory. Even with 5 data points per request, that's 200K metric fetches, costing ~$2 per query. At that scale, CloudWatch costs and API limits are a far bigger concern than a ~100 MB memory difference.

pkg/config/config.go
woehrl01: The standard golang lint error that an error text should start with a lowercase letter.

pkg/clients/cloudwatch/v2/client.go
kgeckhart: We run YACE in a stateless multi-tenant fashion so querying a very large amount of metrics in a short period of time often happens. I was concerned about the potential memory overhead for this change and had shared that with Cristian. 

This wasn't the area I was concerned about initially as we keep our length + period setup to only produce a single datapoint as much as possible. I do agree with Cristian, if we stick to only mapping a single datapoint when the setting is disabled it will ensure there's a minimal overhead to those who upgrade to latest without using the feature.

I was primarily concerned about overhead from switching to a single datapoint -> a slice for our use case. I don't think we should do anything about it now. I wrote a `OneOrMany[T]` that benchmarks nicely vs a single entry slice and would PR it separately if needed.

kgeckhart:
@woehrl01 just wanted to check-in to see if this is something you're still interested in continuing. It's a really valuable feature to have and your contribution is greatly appreciated!

woehrl01:
@kgeckhart I'm open doing the minimal changes discussed above. If you expect to change the whole memory stuff, someone else has to continue based in this PR.

If you are fine with those changes, I can allocate some time next week.

ecktom:
First of all, Thanks @woehrl01 for picking this topic up again. 
We've been playing around with this branch in our environments and it does a decent job and allows us to reduce AWS costs significantly. I can't really see any negative things around the memory allocation though but we're not even close to some scale of millions of metrics though. So what's missing here to get this over the line before it's outdated again? ;)

woehrl01:
Just to clarify, I work on contributions like this in my spare time, so I try to keep the scope focused and the time investment reasonable. I’d appreciate it if that could be respected when commenting on PRs.

If you believe the suggested change is necessary, you're absolutely welcome to take it on and follow up with your own PR. That’s the beauty of open source collaboration.

Also, I want to point out that the wording in your comment came across as unnecessarily confrontational. Using phrases like “baffled by the response” doesn't help foster a respectful or constructive discussion. Please keep the tone collaborative. We're all here to improve the project together.

kgeckhart:
@woehrl01 a change was request to the PR and after some back and forth regarding the change and why it was necessary you ultimately rejected to make it stating,

>  If you expect to change the whole memory stuff, someone else has to continue based in this PR.

Based on your latest reply it seems as though your concern with making the requested change is that you might not have the time to do it, is that correct? It's your work up to this point and I think it's important you keep your name on it vs someone else copying it to make further changes. If you are able to invest the time necessary to resolve as many of the requested changes as possible that would be helpful to the conversation regarding next steps.

woehrl01:
I don't want to invest more time into modifying this PR. I also don't care if my name is afterwards in any commit history, as it's ultimately based on two other people's ideas as mentioned in the PR body. I only care that this is a useful addition to the project. I consider here two ways, either fork the whole PR and open up a third one. Or create a PR to this PR which adds the changes you want to make. Then I will simply merge that PR, so that this PR is ready to be merged.

kgeckhart:
@woehrl01 I have created a PR to your branch which should resolve all the code review feedback, https://github.com/woehrl01/yet-another-cloudwatch-exporter/pull/1.

Thank you again for your effort on this feature.

pkg/clients/cloudwatch/v2/client.go
woehrl01: @kgeckhart I just blindly merged yesterday your suggested changes, to get this PR of my plate. Still I had a look today, about what you changed around your memory concerns. Looking at the current changes, I don't see any change which would reduce memory usage at all. Despite your suggestion, it still creates a slice everytime and it always creates it with full size for all data points. There is now just added complexity around  passing over a flag, to stop the loop early, which you could argue reduced CPU overhead, but is likely neglectable at the scale where the cloudwatch costs would explode.

Before we are moving into premature optimisations, have you actually measured and proved that this actually is an issue? (see my comments above why I doubt that based on numbers). You mention that you are running this at large scale. Maybe you can run the initial PR, side by side for 1-5% of your workload (depending on the scale) and share some realworld memory impact?

kgeckhart: @woehrl01 thanks for taking a look at it and providing feedback.

> Despite your suggestion, it still creates a slice everytime and it always creates it with full size for all data points.

This was an oversight on my part as I was hastily making the change before I needed to go to an event yesterday 😅. My intention was to make sure to only allocate exactly what is necessary and nothing more which I can do with another minor change


> There is now just added complexity around passing over a flag, to stop the loop early, which you could argue reduced CPU overhead, but is likely neglectable at the scale where the cloudwatch costs would explode.

IMO the added complexity is rather minimal and easily testable which is why I went through with the change. 

> Before we are moving into premature optimisations, have you actually measured and proved that this actually is an issue? (see my comments above why I doubt that based on numbers). You mention that you are running this at large scale. Maybe you can run the initial PR, side by side for 1-5% of your workload (depending on the scale) and share some realworld memory impact?

When Cristian brought it up I didn't quite get it at first because we do run it at scale but we do it in such a way that we should only ever get one data point back. We do this intentionally for performance reasons, why ask for data you won't use? I don't know if it was Cristian's intent but my realization was that introducing this feature should not present a noticeable negative impact on the larger community just by upgrading. This exporter is embedded in to places like https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.cloudwatch/ which I know is used by customers at a scale large enough to incur some rather hefty CloudWatch costs. 

Going from a single data point to a slice presents some memory increase that is unlikely to be noticeable for most. We don't know how many people are setting up their configs with a length that is larger than their period (length > period = more than 1 data point) and to what degree they are doing it (2x, 3x, 10x?). This is part of the complexity of building the configs CloudWatch has some incredibly odd behaviors for different metrics so the configs get cargo culted in a way that is not optimal but works. 

Could this guard be unnecessary, yes but the complexity of adding it feels acceptable as a means to try to provide a stable experience to existing users.

Since I messed up the DCO + need another change we will have to figure out what to do with this PR. But first it would probably be good to have @cristiangreco take a look to make sure there's nothing further.

ecktom: @kgeckhart @cristiangreco what's the plan with this PR? 
Don't get me wrong, I understand we're all busy and this is OSS after all. So no pressure here please... ;) I would just highly appreciate some short indication of you guys if and when you plan to move ahead with this PR to be able to make plans on our end as well.

From what I see, this PR seems to be waiting on some action from your side and there is nothing really left the community could support with? If that is wrong and there is still something todo, please let me know, I'm happy to contribute as well.

cristiangreco (approved):
Thanks @woehrl01 and @kgeckhart for your great work here. I feel at this point it's worth merging this work as-is and move on from there.

PR #1657, opened 2025-03-11
prefix with `_` if name begins with a number

shubhindia:


pkg/promutil/prometheus.go
```diff
@@ -194,6 +194,12 @@ func PromStringTag(text string, labelsSnakeCase bool) (bool, string) {
 
 // sanitize replaces some invalid chars with an underscore
 func sanitize(text string) string {
+
+	// metrics starting with a digit violate the prometheus metric naming convention, so we add an underscore
+	if len(text) > 0 && text[0] >= '0' && text[0] <= '9' {
```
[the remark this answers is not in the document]
shubhindia: Yup. Replaced with. regex

pkg/promutil/prometheus.go
```diff
@@ -194,6 +195,12 @@ func PromStringTag(text string, labelsSnakeCase bool) (bool, string) {
 
 // sanitize replaces some invalid chars with an underscore
 func sanitize(text string) string {
+
+	// metrics starting with a digit violate the prometheus metric naming convention, so we add an underscore
+	if matched, _ := regexp.MatchString(`^\d`, text); matched {
```
SuperQ, line 200: Please create a package level var with [`regexp.MustCompile()`](https://pkg.go.dev/regexp#MustCompile) for performance reasons. With this, we take the `regexp.Compile()` performance hit every call.

```go
var metricNumberPrefixRE = regexp.MustCompile(`^\d`)
```

```suggestion
	if matched, _ := metricNumberPrefixRE.MatchString(`^\d`, text); matched {
```

SuperQ (changes requested):

pkg/promutil/prometheus.go
shubhindia: Yup. Addressed

SuperQ (approved):

SuperQ:
It would be useful to add tests for this.

cristiangreco (changes requested):
Can you describe what is the use case here? Metric names are normally prefixed by `aws_<namespace>_` by yace.

PR #1661, opened 2025-03-25
add support for network monitor

tmatias:


pkg/config/services.go
```diff
@@ -686,6 +686,18 @@ var SupportedServices = serviceConfigs{
 			regexp.MustCompile("firewall/(?P<FirewallName>[^/]+)"),
 		},
 	},
+	{
+		Namespace: "AWS/NetworkMonitor",
+		Alias: "networkmonitor",
```
tmatias, line 691: I can use some help to validate this: what is the official alias? I'm setting if to the service prefix string, not completely sure that is the case

cristiangreco: I think this is fine, alias is just a yace specific thing.

pkg/config/services.go
```diff
@@ -686,6 +686,18 @@ var SupportedServices = serviceConfigs{
...
+		ResourceFilters: []*string{
+			aws.String("networkmonitor:monitor"),
+			aws.String("networkmonitor:probe"),
+		},
+		DimensionRegexps: []*regexp.Regexp{
+			regexp.MustCompile("monitor/(?P<MonitorName>[^/]+)"),
+			regexp.MustCompile("probe/(?P<ProbeId>[^/]+)"),
+		},
```
cristiangreco, line 699: Can you please add a test in the aggregator package?

PR #1672, opened 2025-04-07
Run tests in CI and fix failing tests.

jeschkies:
The unit tests weren't running with each PR. This change adds a `test_go` step to the GitHub actions workflow.

Some tests were failing. This is due to a breaking change in Prometheus supporting UTF-8 characters. 

`NameValidationScheme` is set to `LegacyValidation` globally to ensure the behaviour or former releases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,17 @@ on:
   workflow_call:
 
 jobs:
+  test_go:
+    name: Go tests
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/prometheus/golang-builder:1.23-base
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: prometheus/promci@c3c93a50d581b928af720f0134b2b2dad32a6c41 # v0.4.6
+      - uses: ./.github/promci/actions/setup_environment
+      - run: make test
+
   build:
     name: Build for common architectures
     runs-on: ubuntu-latest
@@ -61,7 +72,7 @@ jobs:
   publish_master:
     name: Publish master branch artifacts
     runs-on: ubuntu-latest
-    needs: [build_all, verify-example-configs]
+    needs: [test_go, build_all, verify-example-configs]
     if: |
       (github.repository == 'prometheus-community/yet-another-cloudwatch-exporter')
       &&
@@ -81,7 +92,7 @@ jobs:
   publish_release:
     name: Publish release artifacts
     runs-on: ubuntu-latest
-    needs: [build_all, verify-example-configs]
+    needs: [test_go, build_all, verify-example-configs]
     if: |
       (github.repository == 'prometheus-community/yet-another-cloudwatch-exporter')
       &&
--- a/cmd/yace/scraper.go
+++ b/cmd/yace/scraper.go
@@ -27,7 +27,7 @@ import (
 	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/model"
 )
 
-type scraper struct {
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -10,6 +10,17 @@ on:
...
+  test_go:
+    name: Go tests
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/prometheus/golang-builder:1.23-base
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: prometheus/promci@c3c93a50d581b928af720f0134b2b2dad32a6c41 # v0.4.6
+      - uses: ./.github/promci/actions/setup_environment
+      - run: go test ./pkg/...
```
cristiangreco, line 22: Maybe just `make test` could do?

cristiangreco: We should also extend the `needs` in the `publish_{master,release}` steps later on to depend on this job.

cmd/yace/scraper.go
```diff
@@ -38,16 +38,16 @@ type cachingFactory interface {
 	Clear()
 }
 
-func NewScraper(featureFlags []string) *scraper { //nolint:revive
```
cristiangreco, line 41: Is it worth making scraper public as it's used only in `main.go`?

pkg/promutil/prometheus_test.go
```diff
@@ -71,6 +72,12 @@ func TestSanitize(t *testing.T) {
 }
 
 func TestPromStringTag(t *testing.T) {
+	originalValidationScheme := model.NameValidationScheme
+	model.NameValidationScheme = model.LegacyValidation
+	defer func() {
+		model.NameValidationScheme = originalValidationScheme
+	}()
+
```
cristiangreco, line 80: I think this should raise attention to the breaking changes that added utf8 support:
- https://github.com/prometheus/common/releases/tag/v0.62.0
- https://github.com/prometheus/client_golang/releases/tag/v1.21.0

These two dependencies are now merged on `master` (unreleased), meaning that yace can now potentially export labels with utf8 characters. If the remote write destination does not support utf8, this would result in breaking behaviour right? I'm wondering if we should stick to the old validation scheme for some more time and add a feature flag to enable the new scheme.

cmd/yace/scraper.go
jeschkies: The linter was complaining about the lint being present.

pkg/promutil/prometheus_test.go
jeschkies: I agree. Could we do this in a follow up?

cmd/yace/scraper.go
SuperQ: I thought I saw a nolintlint false-positive bug recently.

jeschkies: That could be it.

.github/workflows/ci.yml
cristiangreco: > We should also extend the `needs` in the `publish_{master,release}` steps later on to depend on this job.

@jeschkies can you please address this other comment?

cmd/yace/scraper.go
cristiangreco: IIUC we could revert this specific change if it was just a false positive?

pkg/promutil/prometheus_test.go
cristiangreco: Can we do the part where we set the legacy validation in this PR? Once pandora's box is opened, I guess it's worth merging a fully usable solution.

cristiangreco (approved):

PR #1690, opened 2025-05-23
Add ElasticBeanstalk DimensionRegexp

benbridts:
- **Add DimensionRegexp to ElasticBeanstalk**
  - We were missing tags on our metrics
- **chore: fix yamllint errors**
  - `make` did complain about this file

### DIFF
--- a/pkg/config/services.go
+++ b/pkg/config/services.go
@@ -211,6 +211,10 @@ var SupportedServices = serviceConfigs{
 		ResourceFilters: []*string{
 			aws.String("elasticbeanstalk:environment"),
 		},
+		DimensionRegexps: []*regexp.Regexp{
+			// arn uses /${ApplicationName}/${EnvironmentName}, but only EnvironmentName is a Metric Dimension
+			regexp.MustCompile("environment/[^/]+/(?P<EnvironmentName>[^/]+)"),
+		},
 	},
 	{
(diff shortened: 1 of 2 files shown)


pkg/config/services.go
```diff
@@ -211,6 +211,10 @@ var SupportedServices = serviceConfigs{
 		ResourceFilters: []*string{
 			aws.String("elasticbeanstalk:environment"),
 		},
+		DimensionRegexps: []*regexp.Regexp{
+			// arn uses /${ApplicationName}/${EnvironmentName}, but only EnvironmentName is a Metric Dimension
+			regexp.MustCompile("environment/[^/]+/(?P<EnvironmentName>[^/]+)"),
```
SuperQ, line 216: I wonder if we should have a policy about naming conventions here. Should we simplify or keep dimension naming exactly the same.

CC @cristiangreco

benbridts: In general I would recommend keeping the same name as the AWS (Console), so that you can easily move between interaces
This also avoids needing to map every new service / metric to a naming convention.

SuperQ (approved):

pkg/config/services.go
cristiangreco: Sorry I missed the conversation, thanks both - LGTM.

PR #1707, opened 2025-07-01
Fix CachingFactory concurrent usage issues

andriikushch:
This PR addresses issues that arise from concurrent usage of the CachingFactory (v1/v2), specifically, concurrent access to the `refreshed` and `cleared` fields outside the mutex lock.

I have added a pair of tests for both versions (v1,v2) to demonstrate these issues.

```sh 
go test -race -run TestRaceConditionRefreshClear .
```

The proposed solution consists of two changes:

1. Adding a mutex lock to the `v1.CachingFactory.Clear` method
2. Change `refreshed` and `cleared` types from `bool` to `go.uber.org/atomic.Bool`

### DIFF
--- a/pkg/clients/v1/factory.go
+++ b/pkg/clients/v1/factory.go
@@ -183,14 +183,14 @@ func NewFactory(logger *slog.Logger, jobsCfg model.JobsConfig, fips bool) *Cachi
 }
 
 func (c *CachingFactory) Clear() {
-	if c.cleared.CompareAndSwap(true, true) {
+	if c.cleared.Load() {
 		return
 	}
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	if c.cleared.CompareAndSwap(true, true) {
+	if c.cleared.Load() {
 		return
 	}
 
@@ -215,14 +215,14 @@ func (c *CachingFactory) Clear() {
 }
 
 func (c *CachingFactory) Refresh() {
-	if c.refreshed.CompareAndSwap(true, true) {
+	if c.refreshed.Load() {
 		return
 	}
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	// Double check Refresh wasn't called concurrently
-	if c.refreshed.CompareAndSwap(true, true) {
+	if c.refreshed.Load() {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/clients/v1/factory.go
```diff
@@ -175,16 +176,21 @@ func NewFactory(logger *slog.Logger, jobsCfg model.JobsConfig, fips bool) *Cachi
...
 func (c *CachingFactory) Clear() {
-	if c.cleared {
+	if c.cleared.CompareAndSwap(true, true) {
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.cleared.CompareAndSwap(true, true) {
```
cristiangreco: @andriikushch I think there's a rebase issue here?

cristiangreco: Right, I see we're introducing a double-check pattern, which was not present in `v1` client before.

kgeckhart: 🤔 is this doing what we expect edit: specifically checking if this is true I thought CAS returned if the value was swapped does it return true even if no actual swap was necessary? It's taking me awhile to think through the semantics of it. I feel like, 

```suggestion
	if c.cleared.Load() {
		return
	}

	c.mu.Lock()
	defer c.mu.Unlock()

	if c.cleared.Load() {
```

should accomplish the same thing without as much complexity.

pkg/clients/v1/factory.go
```diff
@@ -282,7 +288,7 @@ func createAccountClient(logger *slog.Logger, sts stsiface.STSAPI, iam iamiface.
 }
 
 func (c *CachingFactory) GetCloudwatchClient(region string, role model.Role, concurrency cloudwatch_client.ConcurrencyConfig) cloudwatch_client.Client {
-	if !c.refreshed {
+	if c.refreshed.CompareAndSwap(false, false) {
```
kgeckhart: Same comment here regarding usage of CAS vs `!refreshed.Load()`

andriikushch: Good point @kgeckhart. 👍  I will update all the places where `CompareAndSwap` is used.

PR #1729, opened 2025-08-22
Simplify metric to tagged resource join.

jeschkies:


pkg/job/maxdimassociator/associator.go
```diff
@@ -15,80 +15,104 @@ package maxdimassociator
...
-	// mappings is a slice of dimensions-based mappings, one for each regex of a given namespace
-	mappings []*dimensionsRegexpMapping
+	// mapping from join key to tagged resource. The join key is build using the namespace and the required dimensions of the metric.
+	mappings map[uint64]*model.TaggedResource
 
 	logger       *slog.Logger
 	debugEnabled bool
+
+	// mapping from namespace to diffrent join key hash functions for a metric. The functions are sorted by the number of dimensions.
+	namespaceToJoinKeyHashFunction map[string][]hashKeyFunction
```
kgeckhart, line 45: The associator is not intended to be used across different AWS Namespaces. It is always run within the context of a single DiscoveryJob which has a single namespace, https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/15f1631ed924b9c38a7af9f01d2072784d7cd0f9/pkg/job/discovery.go#L90-L96

PR #1760, opened 2025-10-08
Add support for `AWS/EKS` namespace

LS80:
It provide metrics such as `scheduler_pending_pods`.

### DIFF
--- a/pkg/config/services.go
+++ b/pkg/config/services.go
@@ -431,6 +431,16 @@ var SupportedServices = serviceConfigs{
 		},
 	},
 	{
+		Namespace: "AWS/EKS",
+		Alias:     "eks",
+		ResourceFilters: []*string{
+			aws.String("eks:cluster"),
+		},
+		DimensionRegexps: []*regexp.Regexp{
+			regexp.MustCompile(":cluster/(?P<ClusterName>[^/]+)$"),
+		},
+	},
+	{
 		Namespace: "AWS/ELB",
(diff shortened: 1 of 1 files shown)


pkg/config/services.go
```diff
@@ -430,6 +430,16 @@ var SupportedServices = serviceConfigs{
...
+	{
+		Namespace: "AWS/EKS",
+		Alias:     "eks",
+		ResourceFilters: []*string{
+			aws.String("eks:cluster"),
+		},
+		DimensionRegexps: []*regexp.Regexp{
+			regexp.MustCompile(":cluster/(?P<ClusterName>[^/]+)$"),
+		},
+	},
```
cristiangreco, line 442: What's the difference with `ContainerInsights` a few lines above?

LS80: `ContainerInsights` is a namespace for containerized application and worker node metrics that is not available by default for EKS clusters, as it requires an add-on to be installed.

[`AWS/EKS`](https://docs.aws.amazon.com/eks/latest/userguide/cloudwatch.html#cloudwatch-basic-metrics) is a standard namespace that provides metrics from a Kubernetes cluster control plane.

LS80: Ideally all the namespaces listed [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/aws-services-cloudwatch-metrics.html) should be supported?

PR #1790, opened 2025-12-08
Remove unused scraper

jeschkies:
This removes unused code.


andriikushch:
Hi @jeschkies 👋 

This code was included in the last [release](https://github.com/prometheus-community/yet-another-cloudwatch-exporter/tree/v0.63.0).

This PR will remove the mentioned methods from the library API; therefore, it will be a breaking change.

It is possible to check whether these methods/types are used in this project; however, some of them are exported so they can be used programmatically (as a library) by other projects.

At the same time, there is no major version release yet, so we can consider this change if we have a good reason to remove this functionality.

@cristiangreco, what is your opinion on that?

PR #1791, opened 2025-12-08
Test exporter with mocked clients

jeschkies:
This change introduces and end-to-end test for the exporter using AWS client mocks.

PR #1795, opened 2025-12-15
Enhanced metrics

andriikushch:
# Enhanced metrics

This PR adds a new functionality to YACE. It will allow us to get enhanced metrics (EM) for the AWS Services. Enhanced metrics in this scope are metrics that CloudWatch does not return, but can be created using data from other APIs.

One of the goals of this PR is to add this new functionality without introducing any breaking changes to the YACE, neither in the configuration nor in the code, from an API point of view.

Almost all the new code resides in the “internal” package for now. I expect changes and adjustments once it is used in real environments and in different scenarios. Keeping it inside the “internal” package for now will allow changes without breaking the API.

Another goal behind this solution is to make it easy to extend and add additional enhanced metrics if needed.

This PR adds four new metrics: one for each of the following namespaces:

- AWS/Lambda (Timeout)
- AWS/DynamoDB (ItemCount)
- AWS/RDS (AllocatedStorage)
- AWS/ElastiCache (NumCacheNodes)

Important note: it is implemented only for the discovery jobs.

## New components

Screenshot 2026-01-09 at 10 34 22

This solution consists of the following main components:

- Services (service-specific components that return enhanced metrics)
- Clients
- Registry
- enhancedmetrics.Service (component that orchestrates service-specific components)

### Services

Service defines available enhanced metrics for the namespace and their generation.
It has two main methods in its interface: 

- **GetMetrics** - this method must return the requested enhanced metrics. 

Currently, it is not required; However the service should implement the **ListRequiredPermissions** method to provide information about any additional permissions required to call the APIs it uses. It is not required yet, but it may change in the future.

In the given implementations, services are stateless.

### Clients

Services use clients to call the AWS API. Clients may wrap the SDK and provide a more straightforward interface, hiding the complexity of pagination, data translation, and related features. Each service implementation can define its own approach to working with clients, or skip this layer of abstraction entirely for a good reason.

### Registry

The registry acts as a sort of “Factory” for the services. It allows registering services in the runtime for namespaces and, in this way, extends the list of supported enhanced metrics.

This might be an easy way for the developers to create their own exporters using the YACE as a library once this API is exposed.

A service must implement the **MetricsService** interface to be registered in the registry. Currently, there are two methods.

- **GetNamespace** - returns the namespaces to be registered. Each new registration for the same namespace will overwrite the previous one.
- **Instance** - this method should return a new("fresh") instance of the service that should be safe to be used during the scrape operation.

At the moment, this implementation uses the global variable **DefaultEnhancedMetricsService**. It allows for easy addition of new functionality without any breaking changes to the existing library API. In the future, it might be changed and replaced with dependency injection.

### Enhanced metrics service

Enhanced metric service is responsible for orchestrating services within a single scrape job execution. It requires a client factory that implements the RegionalConfigProvider interface to work, performs resource filtering, and delegates work to service-specific services.

## Updated components

- The job model is extended to carry the information about the requested enhanced metrics.
- V2 client’s factory implements the RegionalConfigProvider interface.
- ScrapeAwsData methods create an enhanced processor object and pass it to the runDiscoveryJob method.
- The runDiscoveryJob method uses an enhanced metric processor when needed.

### Known limitation

Added services do not load data concurrently; this was intentionally done in that PR to simplify initial implementation. However, it can be easily changed in the upcoming updates.   

### Testing

Config example:

```yaml
apiVersion: v1alpha1
sts-region: us-east-2
discovery:
  jobs:
    - type: AWS/Lambda
      regions: [ us-east-2 ]
      enhancedMetrics:
        - name: Timeout
    - type: AWS/RDS
      regions: [ us-east-2 ]
      enhancedMetrics:
        - name: AllocatedStorage
    - type: AWS/DynamoDB
      regions: [ us-east-2 ]
      enhancedMetrics:
        - name: ItemCount
    - type: AWS/ElastiCache
      regions: [ us-east-2 ]
      enhancedMetrics:
        - name: NumCacheNodes
```


pkg/internal/enhancedmetrics/service/dynamodb/client.go
```diff
@@ -0,0 +1,104 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to describe table %s: %w", tableName, err)
+	}
+
+	return result.Table, nil
+}
+
+// DescribeAllTables retrieves all DynamoDB tables with their descriptions
+func (c *AWSDynamoDBClient) DescribeAllTables(ctx context.Context, logger *slog.Logger) ([]types.TableDescription, error) {
```
tristanburgess: I'm wondering if code architecture wise it's going to make more sense to have the clients more tightly define a slice of what data should be returned. I'm thinking in terms of how we're probably only at most going to utilize a handful of fields out of the entire resource structs, and yet here we're building a potentially large in memory slice.

pkg/job/internal/enhanced_metrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,241 @@
...
+type Client interface {
+	DescribeAllTables(ctx context.Context, logger *slog.Logger) ([]types.TableDescription, error)
+}
+
+type buildDynamoDBMetricFunc func(context.Context, *slog.Logger, *model.TaggedResource, *types.TableDescription, []string) ([]*model.CloudwatchData, error)
+
+type DynamoDB struct {
+	regionalClients     map[string]Client
+	clientsM            sync.RWMutex
+	clientFactoryMethod func(cfg aws.Config) Client
```
tristanburgess: I think this should be called something like `buildClientFunc`

pkg/job/internal/enhanced_metrics/enhanced_processor.go
```diff
@@ -0,0 +1,51 @@
+package enhanced_metrics
```
tristanburgess: Can package names have `_` now?

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,241 @@
...
+
+	for _, resource := range resources {
+		table, exists := s.regionalData[resource.ARN]
+		if !exists {
+			logger.Warn("dynamodb table not found in data", "arn", resource.ARN)
+			continue
+		}
+
+		for _, enhancedMetric := range enhancedMetrics {
+			if !s.isMetricSupported(enhancedMetric.Name) {
```
tristanburgess: We should probably filter out unsupported metrics separately, otherwise this would do work + log lines for an unsupported metric for every single resource discovered

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,241 @@
...
+		clientFactoryMethod = NewDynamoDBClientWithConfig
+	}
+	svc := &DynamoDB{
+		clientFactoryMethod: clientFactoryMethod,
+		regionalClients:     make(map[string]Client),
+		regionalData:        make(map[string]*types.TableDescription),
+	}
+
+	svc.supportedMetrics = map[string]buildDynamoDBMetricFunc{
+		"ItemCount": svc.buildItemCountMetric,
```
tristanburgess: like maybe we instead associate these metric names with a struct containing the build function as well a list of fields to extract from the AWS API data? Then we only get the data we need to supply the supported metrics.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,241 @@
...
+			}
+
+			result = append(result, &model.CloudwatchData{
+				MetricName:   "ItemCount",
+				ResourceName: resource.ARN,
+				Namespace:    "AWS/DynamoDB",
+				Dimensions:   secondaryIndexesDimensions,
+				Tags:         resource.MetricTags(exportedTags),
+				GetMetricDataResult: &model.GetMetricDataResult{
+					Statistic: "Sum",
```
tristanburgess: I wonder if we'll need a way to not include aggregation information here. It seems a little misleading in terms of CloudWatch semantics because we are not summing/aggregating multiple snapshots of data points over a time period. These are instead true gauges. I think the most accurate way to convey these metrics would be without any statistical modifiers in the resulting metric name

pkg/job/internal/enhanced_metrics/service/dynamodb/service.go
andriikushch: Good point 👍 Let's rename it

pkg/job/internal/enhanced_metrics/enhanced_processor.go
andriikushch: I would say we'd better change the name :)

pkg/internal/enhancedmetrics/service/dynamodb/service.go
tristanburgess: From the meeting: Keep the sum statistic, but add another data point for SampleCount with a hardcoded value of 1.0, this will convey that there's only a single data point per scrape.

andriikushch: I have removed the statistics setting here, as they are "artificial" and don't represent "real" statistics like those for CloudWatch data.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: Good point 👍

pkg/internal/enhancedmetrics/service/dynamodb/client.go
andriikushch: It is a good point, and it might be a nice improvement for the next iteration.

I suggest we do it once the system is more established, including the list of enhanced metrics.

Right now, to add a new metric, you need to define an extra function, and you already have all the data in place (of course, only if it's in the response object). Otherwise, we would need to change the client to map the required fields, and at least change the mapping.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: In this solution, the steps for retrieving data and creating a metric are separated intentionally, so we can start pulling the data earlier.

Therefore, I think this idea matches what we are discussing above https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1795#discussion_r2620549636

pkg/model/model.go
```diff
@@ -43,6 +43,17 @@ type DiscoveryJob struct {
 	ExportedTagsOnMetrics       []string
 	IncludeContextOnInfoMetrics bool
 	DimensionsRegexps           []DimensionsRegexp
+
+	// EnhancedMetrics holds configuration for enhanced metrics in discovery jobs. It contains a configuration for the non-CloudWatch metrics to collect.
+	EnhancedMetrics []*EnhancedMetricConfig
+}
+
+func (d *DiscoveryJob) IsEnhancedMetricsConfigured() bool {
```
tristanburgess: Maybe `HasEnhancedMetrics`?

pkg/job/scrape.go
```diff
@@ -38,6 +39,12 @@ func ScrapeAwsData(
 	awsInfoData := make([]model.TaggedResourceResult, 0)
 	var wg sync.WaitGroup
 
+	enhancedProcessor, err := enhancedmetrics.NewProcessor(factory)
```
tristanburgess: wdyt about making this a var and saving the initialization till the first time we see a discovery job has enhanced metrics configured? then we can return an error if the processor fails to initialize instead of debug logging?

tristanburgess: Also I would prefer the name to be `enhancedMetricsProcessor`

pkg/job/discovery.go
```diff
@@ -62,15 +64,27 @@ func runDiscoveryJob(
...
+	} else {
 		logger.Info("No metrics data found")
-		return resources, nil
 	}
 
-	getMetricDatas, err = gmdProcessor.Run(ctx, svc.Namespace, getMetricDatas)
-	if err != nil {
-		logger.Error("Failed to get metric data", "err", err)
-		return nil, nil
+	// Fetch enhanced metrics if configured
```
tristanburgess: I would remove this comment

pkg/job/discovery.go
```diff
@@ -62,15 +64,27 @@ func runDiscoveryJob(
...
+	if enhancedProcessor != nil && len(job.EnhancedMetrics) > 0 {
+		logger.Debug("Fetching enhanced metrics", "count", len(job.EnhancedMetrics))
+		enhancedData, err := enhancedProcessor.Process(ctx, logger, svc.Namespace, resources, job.EnhancedMetrics, job.ExportedTagsOnMetrics)
+		if err != nil {
+			logger.Warn("Failed to get enhanced metrics", "err", err)
+		} else if len(enhancedData) > 0 {
+			logger.Debug("Got enhanced metrics", "count", len(enhancedData))
```
tristanburgess: Maybe we should say something about the underlying context, like the namespace at least?

pkg/internal/enhancedmetrics/registry.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/service/lambda"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/service/rds"
+)
+
+// DefaultRegistry is the default registry containing all built-in enhanced metrics services
+// It allows registering additional services if needed, or replacing existing ones, e.g. for testing purposes.
+//
+// Note:In the future, it can be removed in favor of being injected via dependency injection.
+// However, it will require changes in the YACE's API.
+var DefaultRegistry = (&Registry{}).
```
tristanburgess: I feel like the naming overlap with Prometheus registries is going to be confusing, maybe we should call it EnhancedMetricServiceRegistry to be more clear?

pkg/internal/enhancedmetrics/processor.go
```diff
@@ -0,0 +1,109 @@
...
+	GetEnhancedMetricsService(namespace string) (service.MetricsService, error)
+}
+
+// Processor is responsible for processing enhanced metrics using appropriate services. It manages multiple enhanced metrics services for different namespaces.
+// It ensures that each service is initialized only once and provides thread-safe access to these services.
+// The Processor uses a RegionalConfigProvider to obtain AWS configurations for different regions and roles.
+// It is intended to be used in the scope of a single scrape operation.
+type Processor struct {
+	ConfigProvider          config.RegionalConfigProvider
+	EnhancedMetricsServices map[string]service.MetricsService
```
tristanburgess: wdyt about renaming the type `MetricsService` to `EnhancedMetricsService`?

pkg/internal/enhancedmetrics/service.go
```diff
@@ -0,0 +1,109 @@
...
+	"sync"
+
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/clients"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/config"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/service"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/model"
+)
+
+// MetricsServiceRegistry defines an interface to get enhanced metrics services by namespace
+type MetricsServiceRegistry interface {
```
tristanburgess, line 27: Similar comment as before, rename to `EnhancedMetricsServiceRegistry`

pkg/internal/enhancedmetrics/service/services.go
```diff
@@ -0,0 +1,32 @@
...
+	"context"
+	"log/slog"
+
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/config"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/model"
+)
+
+// MetricsService is the interface that enhanced metrics services should implement to be used by the enhanced metrics processor.
+type MetricsService interface {
+	// LoadMetricsMetadata should load any metadata needed for the enhanced metrics service. It should be concurrent safe.
```
tristanburgess: It's not clear looking at this interface what this metadata should look like or how it's intended to be used. The function signature only returns an error, so where is this metadata supposed to be loaded?

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (s *DynamoDB) ListRequiredPermissions() []string {
+	return []string{
+		"dynamodb:DescribeTable",
+		"dynamodb:ListTables",
+	}
+}
+
+func (s *DynamoDB) ListSupportedMetrics() []string {
```
tristanburgess: I think we might want to expose these as well in the YACE service map, i.e. in https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/c8586899a1d4696b8ad54b13271c499ed8ace8e1/pkg/config/services.go#L85

The use case is that I need a way to expose these supported enhanced metrics so I can easily loop over all supported services and collect them for each service.

what do you think?

tristanburgess: I pushed a commit which adds this, though I'm not sure that we want to settle on including the internal packages in the services.go. Open to ideas

https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1795/changes/e3abd320e762bfaf05dc481107f814f12c3795d0

pkg/job/scrape.go
```diff
@@ -38,6 +39,12 @@ func ScrapeAwsData(
...
+	if err != nil {
+		logger.Debug("Couldn't initialize enhanced metrics processor", "err", err)
```
tristanburgess: Do you think this should be a Warning level message? Or is it expected that this fails in most cases?

pkg/job/discovery.go
andriikushch: removed

pkg/internal/enhancedmetrics/registry.go
andriikushch: I have renamed it to `DefaultEnhancedMetricServiceRegistry`. I would like to keep "Default" in the name to emphasize that it is not a single possible instance of the registry.

In the current PR, it makes sense only for testing purposes; however, in the future, custom registries could be created when YACE is used as a library and used to configure enhanced metrics.

pkg/internal/enhancedmetrics/processor.go
andriikushch: Let's do it 👍 I think it is a good idea as the package name doesn't provide enough context in this case.

pkg/internal/enhancedmetrics/service.go
andriikushch: In this case, I think we can keep the name as it is, because the package name is `enhancedmetrics`. So it will be `enhancedmetrics.MetricsServiceRegistry`.

pkg/internal/enhancedmetrics/service/services.go
andriikushch: Good point, I will add a better description.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: I would consider extending `type MetricsService interface` to have `ListSupportedEnhancedMetrics() []string` method, then we can expose all this data from the registry (we might need to extend `type MetricsServiceRegistry interface` too).

In this case, we could generalize the code here and avoid importing each service separately.

Speaking about the change to the `SupportedServices`, it is a change to the YACE API, I assume you would like to use this functionality in the code (as a library); however, the idea of using the internal package is to prevent this for now (until interfaces are finally defined and the feature is stable).

I would suggest, if it is possible, hardcoding these values on your side for now.

pkg/job/scrape.go
andriikushch: At the moment, it will fail for all the cases where the factory is not implementing `RegionalConfigProvider`, in other words, all existing usages of the YACE as a library. Therefore, I think we can keep it as debug, at least for the initial release.

pkg/config/services.go
```diff
@@ -352,6 +372,7 @@ var SupportedServices = serviceConfigs{
 			regexp.MustCompile("cluster:(?P<CacheClusterId>[^/]+)"),
 			regexp.MustCompile("serverlesscache:(?P<clusterId>[^/]+)"),
 		},
+		AvailableEnhancedMetrics: elasticache.NewElastiCacheService(nil).ListSupportedEnhancedMetrics(),
```
jeschkies: It feels like this should work without constructing the service.

pkg/exporter_test.go
```diff
@@ -0,0 +1,598 @@
+package exporter
```
jeschkies, line 13: It's probably easier to review if land https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1791 first.

pkg/internal/enhancedmetrics/clients/clients.go
```diff
@@ -0,0 +1,90 @@
...
+	"sync"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/config"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/model"
+)
+
+// separator is NULL byte (0x00)
+const separator = "\x00"
```
jeschkies: Where is that used? Maybe add it do the documentation.

pkg/internal/enhancedmetrics/clients/clients.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Clients manages AWS service clients per region and role.
+// It is used by enhanced metrics services and allows them to initialize a client only once per region and role and then reuse it.
+type Clients[T any] struct {
```
jeschkies: It feels this is more like a client cache. What am I missing?

Also, could you narrow `T`?

pkg/internal/enhancedmetrics/clients/clients.go
```diff
@@ -0,0 +1,90 @@
...
+	regionalClients map[string]T
+	clientsM        sync.RWMutex
+
+	// buildClientFunc is a function that builds a client of type T given an AWS config.
+	buildClientFunc func(cfg aws.Config) T
```
jeschkies: So `Clients` is actually a factory isn't it?

pkg/job/scrape.go
```diff
@@ -52,14 +59,22 @@ func ScrapeAwsData(
 					}
 					jobLogger = jobLogger.With("account", accountID)
 
+					// at this point we have already all the data to start loading the enhanced metrics if any is configured
+					if discoveryJob.IsEnhancedMetricsConfigured() && enhancedProcessor != nil {
+						err := enhancedProcessor.LoadMetricsMetadata(ctx, jobLogger, region, role, discoveryJob.Namespace, enhancedmetrics.DefaultEnhancedMetricServiceRegistry)
+						if err != nil {
```
jeschkies: Why are we loading metadata at this point and not later?

pkg/job/discovery.go
```diff
@@ -43,6 +44,7 @@ func runDiscoveryJob(
 	clientTag tagging.Client,
 	clientCloudwatch cloudwatch.Client,
 	gmdProcessor getMetricDataProcessor,
+	enhancedProcessor *em.Processor,
```
jeschkies: Why not use the interface here?

pkg/job/discovery.go
```diff
@@ -62,15 +64,26 @@ func runDiscoveryJob(
 
 	svc := config.SupportedServices.GetService(job.Namespace)
 	getMetricDatas := getMetricDataForQueries(ctx, logger, job, svc, clientCloudwatch, resources)
-	if len(getMetricDatas) == 0 {
+
+	if len(getMetricDatas) > 0 && svc != nil {
+		getMetricDatas, err = gmdProcessor.Run(ctx, svc.Namespace, getMetricDatas)
+		if err != nil {
+			logger.Error("Failed to get metric data", "err", err)
+			return nil, nil
```
jeschkies: Shouldn't we retrun an error?

pkg/job/discovery.go
```diff
@@ -62,15 +64,26 @@ func runDiscoveryJob(
 
 	svc := config.SupportedServices.GetService(job.Namespace)
 	getMetricDatas := getMetricDataForQueries(ctx, logger, job, svc, clientCloudwatch, resources)
-	if len(getMetricDatas) == 0 {
+
+	if len(getMetricDatas) > 0 && svc != nil {
+		getMetricDatas, err = gmdProcessor.Run(ctx, svc.Namespace, getMetricDatas)
```
jeschkies: Could we rename `getMetricDatas`?

pkg/internal/enhancedmetrics/registry.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/service/elasticache"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/service/lambda"
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/internal/enhancedmetrics/service/rds"
+)
+
+// DefaultEnhancedMetricServiceRegistry is the default registry containing all built-in enhanced metrics services
+// It allows registering additional services if needed, or replacing existing ones, e.g. for testing purposes.
+//
+// Note:In the future, it can be removed in favor of being injected via dependency injection.
+// However, it will require changes in the YACE's API.
```
jeschkies, line 30: Where is this used or required? While I agree that it would be better to have a registry or pass some kind of context the current design already has a global var https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/c8586899a1d4696b8ad54b13271c499ed8ace8e1/pkg/config/services.go#L85 at lest this is how I understand `var SupportedServices`. I'd follow the same design for now and then change it. Changing it might require a new major version though.

pkg/internal/enhancedmetrics/registry.go
```diff
@@ -0,0 +1,74 @@
...
+type MetricsService interface {
+	Instance() service.EnhancedMetricsService
+	GetNamespace() string
+}
+
+// Registry maintains a mapping of enhanced metrics services by their namespaces.
+type Registry struct {
+	m sync.RWMutex
+
+	services map[string]func() service.EnhancedMetricsService
```
jeschkies, line 48: This seems very similar to https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/c8586899a1d4696b8ad54b13271c499ed8ace8e1/pkg/config/services.go#L85

pkg/internal/enhancedmetrics/service/dynamodb/client.go
```diff
@@ -0,0 +1,99 @@
...
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+)
+
+type awsClient interface {
```
jeschkies, line 25: What is the purpose of a private interface?

pkg/internal/enhancedmetrics/service/dynamodb/client.go
```diff
@@ -0,0 +1,99 @@
...
+	ListTables(ctx context.Context, params *dynamodb.ListTablesInput, optFns ...func(*dynamodb.Options)) (*dynamodb.ListTablesOutput, error)
+	DescribeTable(ctx context.Context, params *dynamodb.DescribeTableInput, optFns ...func(*dynamodb.Options)) (*dynamodb.DescribeTableOutput, error)
+}
+
+// AWSDynamoDBClient wraps the AWS DynamoDB client
+type AWSDynamoDBClient struct {
+	client awsClient
+}
```
jeschkies, line 32: That's am interesting desing. Why not use an exported interface and an implementation using a struct?

pkg/model/model.go
andriikushch: Done

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,243 @@
...
+type buildDynamoDBMetricFunc func(context.Context, *slog.Logger, *model.TaggedResource, *types.TableDescription, []string) ([]*model.CloudwatchData, error)
+
+type DynamoDB struct {
+	clients *clients.Clients[Client]
+
+	regionalData map[string]*types.TableDescription
+	dataM        sync.RWMutex
+
+	supportedMetrics map[string]buildDynamoDBMetricFunc
+	buildClientFunc  func(cfg aws.Config) Client
```
jeschkies, line 50: Why is this required?

pkg/job/discovery.go
andriikushch: Done

pkg/job/scrape.go
andriikushch: > Also I would prefer the name to be enhancedMetricsProcessor

Agree, I have changed the name

andriikushch: > wdyt about making this a var and saving the initialization till the first time we see a discovery job has enhanced metrics configured

This is a good idea, we can save some resources there.

> then we can return an error if the processor fails to initialize instead of debug logging

I think we'd better log a warning (instead of debug) and continue without enhanced metrics. What do you think?

pkg/config/services.go
andriikushch: There is a thread for that topic: https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1795#discussion_r2645865641

I think that we can avoid changing this type (service config) for now.

pkg/exporter_test.go
andriikushch: This is a good idea for the regression testing in the first place, and we can reuse some mocks.

We can review and merge it. However, let's not block this one because of it.

cc: @tristanburgess

pkg/internal/enhancedmetrics/clients/clients.go
andriikushch: Good point, I have extended the description here:

```
// This is an arbitrary choice of a character that is unlikely to appear in region names or role ARNs/ExternalIDs.
// It is used to separate different components in the client key in the map when generating unique keys for region-role combinations.
```

pkg/internal/enhancedmetrics/clients/clients.go
andriikushch: > It feels this is more like a client cache. What am I missing?

From one perspective, it can be considered a short-lived cache (with a lifetime of one `ScrapeAwsData` call). This type is used to initialize clients when needed and cache them for reuse.

> Also, could you narrow `T`

It is a good question.

Underlying clients that are supposed to be used with this type don't have a shared interface. They are namespace-specific, for example:

```
type Client interface {
	DescribeAllDBInstances(ctx context.Context, logger *slog.Logger) ([]types.DBInstance, error)
}
```

Also, this implementation doesn't have any specific requirements for `T`, which is why I decided to go with `T`.

One possible option is to use a "marker" interface; however, it is artificial and unclear which problem it is supposed to solve.

pkg/internal/enhancedmetrics/clients/clients.go
andriikushch: Conceptually, yes. 

(I say "conceptually" because the word "factory" might make thinking of the specific pattern, which is not precisely correct here). 

`Clients` is responsible for creating a client instance from the provided config and returning it, or returning an existing one if it has already been initialized.

pkg/internal/enhancedmetrics/service/dynamodb/client.go
andriikushch: These interfaces are private to reduce their visibility, as they are used only for the mocks that test the clients' wrappers defined in that implementation.

Wrapped clients encapsulate logic, such as pagination, and their interfaces are narrower and more semantically meaningful in the context of the enhanced metrics. Therefore, they are used in the enhanced services' tests.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: @tristanburgess, I will remove changes introduced in the https://github.com/prometheus-community/yet-another-cloudwatch-exporter/commit/e3abd320e762bfaf05dc481107f814f12c3795d0 for now.

pkg/internal/enhancedmetrics/service/dynamodb/client.go
andriikushch: The idea is as follows:

1. There is an SDK client awsClient, and there are only two methods that we actually need in this implementation:

```
type awsClient interface {
	ListTables(ctx context.Context, params *dynamodb.ListTablesInput, optFns ...func(*dynamodb.Options)) (*dynamodb.ListTablesOutput, error)
	DescribeTable(ctx context.Context, params *dynamodb.DescribeTableInput, optFns ...func(*dynamodb.Options)) (*dynamodb.DescribeTableOutput, error)
}
```

2. What we actually need, in the service, is to describe all the tables. This will allow access to all the data necessary to build enhanced metrics.

3. To achieve it, we should list all the tables respecting the pagination and call the describe method on each.

4. This logic is abstracted by the implementation of the:

```
type Client interface {
	DescribeAllTables(ctx context.Context, logger *slog.Logger) ([]types.TableDescription, error)
}
```

The ultimate goal of the extra abstraction layer is to encapsulate pagination/data retrieval logic and provide a simple, specific interface.

pkg/config/services.go
tristanburgess: Agreed, please feel free to remove this change and in the future we'll make it part of the registry to create a map of all supported services with their supported enhanced metrics

pkg/job/scrape.go
tristanburgess: Ohh right, yes that makes sense

pkg/internal/enhancedmetrics/service.go
tristanburgess: Ah gotcha, yeah that's a good pattern I think, let's keep it as is

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: This function is used in the `Clients` to create new client instances when they are needed.

pkg/job/scrape.go
tristanburgess: yeah I think this makes sense :+1:

pkg/internal/enhancedmetrics/registry.go
andriikushch: Correct. However, changes in the `SupportedServices` are reverted for now.

pkg/config/services.go
andriikushch: Removed

pkg/internal/enhancedmetrics/registry.go
andriikushch: Having it as a global variable allows a few things:

- Keep the existing API as it is; we might need to change the `ScrapeAwsData` signature.
- Simplify testing; we can replace real services with mocks. It is the primary usecase for it at the moment (given it is in the `internal` folder).
- Also, it is a way to extend functionality by easy registration of the extra services.

In the following iterations, we might consider at least two changes:

- If we proceed with a global variable, we should make it concurrently safe.
- If we switch to the injection, we need to change the API.

Note: given the current version `0.63.0`, we might not need a major version bump to change if we introduce a breaking change. However, we could postpone this decision right now, as we are not introducing new contracts.

pkg/job/discovery.go
andriikushch: I think using an interface here is not a bad idea. 

I didn't do it initially because there was no need. `runDiscoveryJob` isn't exposed, and there is no second implementation of `Processor`.

Let me add one.

pkg/job/discovery.go
andriikushch: Let's do it

pkg/job/discovery.go
andriikushch: `runDiscoveryJob` doesn't return an error in the current implementation, so I keep this behavior. 

Also, if the processing of the CloudWatch metrics failed, we might still have an enhanced metric to process. I think we should not stop here.

pkg/job/scrape.go
andriikushch: The idea is to start loading as soon as possible, to have a room to increase concurrency and reduce scraping time.

To simplify this PR, enhanced metric services load data synchronously; however, each enhanced metric service can decide how to do so. Assuming that loading data is mostly done via http calls, it has the potential to leverage the benefits of concurrent execution.

docs/configuration.md
```diff
@@ -118,6 +118,15 @@ dimensionNameRequirements:
 # useful when cloudwatch metrics might not be present or when using info metrics to understand where your resources exist
 [ includeContextOnInfoMetrics: <boolean> ]
 
+# (optional) This is an experimental feature that can be used to enable enhanced metrics for specific services within this discovery job. It might be subject to changes in future releases.
+# Currently supported enhanced metrics are:
+# - AWS/Lambda (Timeout)
+# - AWS/DynamoDB (ItemCount)
+# - AWS/RDS (AllocatedStorage)
+# - AWS/ElastiCache (NumCacheNodes)
```
jeschkies: What would the values look like?

pkg/internal/enhancedmetrics/processor.go
```diff
@@ -0,0 +1,109 @@
...
+	if !ok {
+		// should not happen because of ensureServiceInitialized call above
+		return fmt.Errorf("enhanced metrics service for namespace %s not initialized", namespace)
+	}
+
+	return svc.LoadMetricsMetadata(ctx, logger, region, role, ep.ConfigProvider)
+}
+
+// Process processes the enhanced metrics for the specified namespace using the appropriate enhanced metrics service.
+func (ep *Processor) Process(ctx context.Context, logger *slog.Logger, namespace string, resources []*model.TaggedResource, metrics []*model.EnhancedMetricConfig, exportedTagOnMetrics []string) ([]*model.CloudwatchData, error) {
```
jeschkies:
```suggestion
func (ep *Processor) Process(ctx context.Context, logger *slog.Logger, namespace string, resources []*model.TaggedResource, metricConfig []*model.EnhancedMetricConfig, exportedTagOnMetrics []string) ([]*model.CloudwatchData, error) {
```

pkg/job/scrape.go
```diff
@@ -52,14 +73,33 @@ func ScrapeAwsData(
 					}
 					jobLogger = jobLogger.With("account", accountID)
 
+					// at this point we have already all the data to start loading the enhanced metrics if any is configured
+					if discoveryJob.HasEnhancedMetrics() && enhancedMetricsProcessor != nil {
+						err := enhancedMetricsProcessor.LoadMetricsMetadata(ctx, jobLogger, region, role, discoveryJob.Namespace, enhancedmetrics.DefaultEnhancedMetricServiceRegistry)
+						if err != nil {
+							jobLogger.Error("Couldn't load enhanced metrics metadata", "err", err)
+						}
+					}
```
jeschkies: I think this should be moved into `runDiscoveryJob` because `LoadMetricsMetadata` is making the actual calls to AWS. And that should habben in `runDiscoveryJob`.

pkg/internal/enhancedmetrics/registry.go
jeschkies: I think I was misunderstood 🙂 I would replace the `var DefaultEnhancedMetricServiceRegistry` with a concrete struct similar to https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/c8586899a1d4696b8ad54b13271c499ed8ace8e1/pkg/config/services.go#L85. Your solution is cleaner but should be a follow up that also changes `SupportedServices`.

pkg/internal/enhancedmetrics/registry.go
jeschkies: > However, changes in the SupportedServices are reverted for now.

Ah. So you did change it?

andriikushch: yes

docs/configuration.md
andriikushch: Good point 👍  
I have added a detailed description of what these values represent; now it should be clear.

pkg/internal/enhancedmetrics/clients/clients.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// NewClients creates a new Clients instance.
+func NewClients[T any](buildClientFunc func(cfg aws.Config) T) *Clients[T] {
```
jeschkies: I think we should restrict `T`.

pkg/internal/enhancedmetrics/processor.go
```diff
@@ -0,0 +1,109 @@
...
+	ep.m.RLock()
+	defer ep.m.RUnlock()
+
+	svc, ok := ep.EnhancedMetricsServices[namespace]
+	if !ok {
+		// should not happen because of ensureServiceInitialized call above
+		return fmt.Errorf("enhanced metrics service for namespace %s not initialized", namespace)
+	}
+
+	return svc.LoadMetricsMetadata(ctx, logger, region, role, ep.ConfigProvider)
```
jeschkies: This makes the services stateful right? I think we shouldn't do that.

pkg/internal/enhancedmetrics/service/elasticache/service.go
```diff
@@ -0,0 +1,209 @@
...
+
+	s.regionalData = make(map[string]*types.CacheCluster)
+
+	instances, err := client.DescribeAllCacheClusters(ctx, logger)
+	if err != nil {
+		return fmt.Errorf("error listing cache clusters in region %s: %w", region, err)
+	}
+
+	for _, instance := range instances {
+		s.regionalData[*instance.ARN] = &instance
```
jeschkies: I would not persist the data here unless there's a cache. Which doesn't work either because the scraping is stateless.

pkg/job/discovery.go
```diff
@@ -61,19 +66,30 @@ func runDiscoveryJob(
...
+	} else {
 		logger.Info("No metrics data found")
-		return resources, nil
 	}
 
-	getMetricDatas, err = gmdProcessor.Run(ctx, svc.Namespace, getMetricDatas)
-	if err != nil {
-		logger.Error("Failed to get metric data", "err", err)
-		return nil, nil
+	if enhancedProcessor != nil && len(job.EnhancedMetrics) > 0 && svc != nil {
```
jeschkies: I think we should call something like `getEnhancedMetricsData` here.

pkg/job/scrape.go
andriikushch: Moving it there

pkg/internal/enhancedmetrics/processor.go
andriikushch: I am working on it.

It will change the architecture; however, if the processor call is moved to the `runDiscoveryJob`, it should fit well.

pkg/internal/enhancedmetrics/service/elasticache/service.go
andriikushch: It won't be required once I change `LoadMetricsMetadata` to return the data.

pkg/internal/enhancedmetrics/processor.go
jeschkies: I think https://go.dev/blog/pipelines explains my point a little better.

pkg/internal/enhancedmetrics/clients/clients.go
andriikushch: I have removed Clients:

1. To make services stateless
2. Even in previous architecture where processors and services were stateful, cache was used only in the scenario when a single request contained multiple jobs configured to scrape the same namespaces. 
3. It simplifies a PR significantly.

pkg/internal/enhancedmetrics/processor.go
andriikushch: It makes sense to switch to the stateless version, at least for now. 

The initial decision stemmed from premature optimization, with the idea of two steps: loading data and processing. It seemed easier to add concurrency and reuse this data, let's say, to get extra labels. However, after recent refactoring, it doesn't look really required at least for the initial version.

Thank you for discussing it 👍

docs/configuration.md
```diff
@@ -118,6 +118,15 @@ dimensionNameRequirements:
 # useful when cloudwatch metrics might not be present or when using info metrics to understand where your resources exist
 [ includeContextOnInfoMetrics: <boolean> ]
 
+# (optional) This is an experimental feature that can be used to enable enhanced metrics for specific services within this discovery job. It might be subject to changes in future releases.
+# Currently supported enhanced metrics are:
+# - AWS/Lambda (Timeout) The maximum execution duration permitted for the function before termination.
+# - AWS/DynamoDB (ItemCount) The count of items in the table, updated approximately every six hours; may not reflect recent changes.
+# - AWS/RDS (AllocatedStorage) The storage capacity in gibibytes (GiB) allocated for the DB instance.
+# - AWS/ElastiCache (NumCacheNodes) The count of cache nodes in the cluster; must be 1 for Valkey or Redis OSS clusters, or between 1 and 40 for Memcached clusters.
```
jeschkies: Please give an example on how this is done.

pkg/internal/enhancedmetrics/service.go
```diff
@@ -0,0 +1,85 @@
...
+	logger *slog.Logger,
+	namespace string,
+	resources []*model.TaggedResource,
+	metrics []*model.EnhancedMetricConfig,
+	exportedTagOnMetrics []string,
+	region string,
+	role model.Role,
+	enhancedMetricsServiceRegistry MetricsServiceRegistry,
+) ([]*model.CloudwatchData, error) {
+	svc, err := enhancedMetricsServiceRegistry.GetEnhancedMetricsService(namespace)
```
jeschkies: How do you feel about moving the registry into the service struct?

pkg/internal/enhancedmetrics/service/dynamodb/client.go
jeschkies: Ah. So the idea is to test `AWSDynamoDBClient.DescribeAllTables` with an AWS client mock, right?

pkg/internal/enhancedmetrics/service/dynamodb/client.go
```diff
@@ -0,0 +1,99 @@
...
+				logger.Error("Failed to describe table", "error", err.Error(), "table", tableName)
+				continue
+			}
+			allTables = append(allTables, *tableDesc)
+		}
+
+		if lastEvaluatedTableName == nil {
+			break
+		}
+		startTableName = lastEvaluatedTableName
```
jeschkies: You don't have to do it now but it seems this could be done with https://github.com/grafana/dskit/blob/main/concurrency/runner.go in a follow up.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
jeschkies: Aren't we using a factory pattern in other places? This is not a blocker but I think it should align.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,240 @@
...
+	svc.supportedMetrics = map[string]buildDynamoDBMetricFunc{
+		// The count of items in the table, updated approximately every six hours; may not reflect recent changes.
+		"ItemCount": svc.buildItemCountMetric,
+	}
+
+	return svc
+}
+
+func (s *DynamoDB) GetNamespace() string {
+	return "AWS/DynamoDB"
```
jeschkies: Could you move this into a constant?

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (s *DynamoDB) isMetricSupported(metricName string) bool {
+	_, exists := s.supportedMetrics[metricName]
+	return exists
+}
+
+func (s *DynamoDB) GetMetrics(
+	ctx context.Context,
+	logger *slog.Logger,
+	namespace string,
```
jeschkies: Why would the service interface take a `namespace` parameter?

pkg/internal/enhancedmetrics/service/dynamodb/service.go
```diff
@@ -0,0 +1,240 @@
...
+		return nil, fmt.Errorf("dynamodb enhanced metrics service cannot process namespace %s", namespace)
+	}
+
+	// filter only supported enhanced metrics
+	var enhancedMetricsFiltered []*model.EnhancedMetricConfig
+	for _, em := range enhancedMetricConfigs {
+		if s.isMetricSupported(em.Name) {
+			enhancedMetricsFiltered = append(enhancedMetricsFiltered, em)
+		} else {
+			logger.Warn("enhanced metric not supported, skipping", "metric", em.Name)
```
jeschkies: Shouldn't this be validated in the config creation?

pkg/internal/enhancedmetrics/service/dynamodb/service_test.go
```diff
@@ -0,0 +1,278 @@
...
+			require.NotNil(t, got)
+			require.Len(t, got.supportedMetrics, 1)
+			require.NotNil(t, got.supportedMetrics["ItemCount"])
+		})
+	}
+}
+
+func TestDynamoDB_GetNamespace(t *testing.T) {
+	service := NewDynamoDBService(nil)
+	expectedNamespace := "AWS/DynamoDB"
```
jeschkies: Use the constant

pkg/internal/enhancedmetrics/service/elasticache/service.go
```diff
@@ -0,0 +1,202 @@
...
+
+	for _, resource := range resources {
+		cluster, exists := data[resource.ARN]
+		if !exists {
+			logger.Warn("elasticache cluster not found in data", "arn", resource.ARN)
+			continue
+		}
+
+		for _, enhancedMetric := range enhancedMetricsFiltered {
+			em, err := s.supportedMetrics[enhancedMetric.Name](ctx, logger, resource, cluster, exportedTagOnMetrics)
```
jeschkies: That's tricky to follow. What's happening here? ... Ok, if I'm not mistaken your are building the Cloudwatch data right?

pkg/internal/enhancedmetrics/service/elasticache/service.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+
+			result = append(result, em)
+		}
+	}
+
+	return result, nil
+}
+
+func (s *ElastiCache) buildNumCacheNodesMetric(_ context.Context, _ *slog.Logger, resource *model.TaggedResource, cluster *types.CacheCluster, exportedTags []string) (*model.CloudwatchData, error) {
```
jeschkies:
```suggestion
func (s *ElastiCache) buildNumCacheNodesMetric(_ context.Context, _ *slog.Logger, resource *model.TaggedResource, cluster *types.CacheCluster, exportedTags []string) (*model.CloudwatchData, error) {
```
```suggestion
func buildNumCacheNodesMetric(_ context.Context, _ *slog.Logger, resource *model.TaggedResource, cluster *types.CacheCluster, exportedTags []string) (*model.CloudwatchData, error) {
```
As far as I can tell this function doesn't use `s`

pkg/internal/enhancedmetrics/service/elasticache/service.go
```diff
@@ -0,0 +1,202 @@
...
+		regionalConfigProvider,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't load elasticache metrics metadata: %w", err)
+	}
+
+	var result []*model.CloudwatchData
+
+	for _, resource := range resources {
+		cluster, exists := data[resource.ARN]
```
jeschkies: Is cluster the elasticache cluster? Could you rename it?

pkg/internal/enhancedmetrics/service/lambda/client.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"log/slog"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/lambda"
+	"github.com/aws/aws-sdk-go-v2/service/lambda/types"
+)
+
+type awsClient interface {
+	ListFunctions(ctx context.Context, params *lambda.ListFunctionsInput, optFns ...func(*lambda.Options)) (*lambda.ListFunctionsOutput, error)
```
jeschkies, line 26: Where is `optFns` used? I cannot find it or is this interface prescribed by the AWS client

pkg/internal/enhancedmetrics/service/lambda/service.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+
+			result = append(result, em)
+		}
+	}
+
+	return result, nil
+}
+
+func (s *Lambda) buildTimeoutMetric(_ context.Context, _ *slog.Logger, resource *model.TaggedResource, fn *types.FunctionConfiguration, exportedTags []string) (*model.CloudwatchData, error) {
```
jeschkies:
```suggestion
func buildTimeoutMetric(_ context.Context, _ *slog.Logger, resource *model.TaggedResource, fn *types.FunctionConfiguration, exportedTags []string) (*model.CloudwatchData, error) {
```

pkg/internal/enhancedmetrics/service/lambda/service.go
```diff
@@ -0,0 +1,195 @@
...
+
+	for _, resource := range resources {
+		fn, exists := data[resource.ARN]
+		if !exists {
+			logger.Warn("Lambda function not found in data", "arn", resource.ARN)
+			continue
+		}
+
+		for _, enhancedMetric := range enhancedMetricsFiltered {
+			em, err := s.supportedMetrics[enhancedMetric.Name](ctx, logger, resource, fn, exportedTagOnMetrics)
```
jeschkies: I'd do something like
```suggestion
			em, err := s.supportedMetrics[enhancedMetric.Name].ConvertMetric(ctx, logger, resource, fn, exportedTagOnMetrics)
```
this makes it easier to follow and understand.

pkg/internal/enhancedmetrics/service/rds/client.go
```diff
@@ -0,0 +1,76 @@
...
+	logger.Debug("Describing all RDS DB instances")
+	var allInstances []types.DBInstance
+	var marker *string
+	maxRecords := aws.Int32(100)
+
+	for {
+		output, err := c.describeDBInstances(ctx, &rds.DescribeDBInstancesInput{
+			Marker:     marker,
+			MaxRecords: maxRecords,
+		})
```
jeschkies, line 67: It seems we could filter by ARNs https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/rds#DescribeDBInstancesInput

pkg/internal/enhancedmetrics/service/rds/service.go
```diff
@@ -0,0 +1,216 @@
...
+		// The storage capacity in gibibytes (GiB) allocated for the DB instance.
+		"AllocatedStorage": rds.buildAllocatedStorageMetric,
+	}
+
+	return rds
+}
+
+// GetNamespace returns the AWS CloudWatch namespace for RDS
+func (s *RDS) GetNamespace() string {
+	return "AWS/RDS"
```
jeschkies: Use a constant

pkg/internal/enhancedmetrics/service/rds/service.go
```diff
@@ -0,0 +1,216 @@
...
+		regionalConfigProvider,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error loading RDS metrics metadata: %w", err)
+	}
+
+	var result []*model.CloudwatchData
+
+	for _, resource := range resources {
+		dbInstance, exists := data[resource.ARN]
```
jeschkies, line 139: Is there a way not to describe all RDS instance but only query for those with certain ARNs?

pkg/internal/enhancedmetrics/service/rds/service.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+
+			result = append(result, em)
+		}
+	}
+
+	return result, nil
+}
+
+func (s *RDS) buildAllocatedStorageMetric(_ context.Context, _ *slog.Logger, resource *model.TaggedResource, instance *types.DBInstance, exportedTags []string) (*model.CloudwatchData, error) {
```
jeschkies:
```suggestion
func buildAllocatedStorageMetric(_ context.Context, _ *slog.Logger, resource *model.TaggedResource, instance *types.DBInstance, exportedTags []string) (*model.CloudwatchData, error) {
```

pkg/internal/enhancedmetrics/service/rds/service.go
```diff
@@ -0,0 +1,216 @@
...
+		ResourceName: resource.ARN,
+		Namespace:    "AWS/RDS",
+		Dimensions:   dimensions,
+		Tags:         resource.MetricTags(exportedTags),
+
+		// Store the value as a single data point
+		GetMetricDataResult: &model.GetMetricDataResult{
+			DataPoints: []model.DataPoint{
+				{
+					Value:     &valueInGiB,
```
jeschkies: Shall we use bytes instead?

pkg/job/discovery.go
```diff
@@ -61,19 +78,44 @@ func runDiscoveryJob(
...
-		return resources, nil
 	}
 
-	getMetricDatas, err = gmdProcessor.Run(ctx, svc.Namespace, getMetricDatas)
+	if enhancedMetricsService == nil || !job.HasEnhancedMetrics() || svc == nil {
+		return resources, metricData
+	}
+
+	logger.Debug("Processing enhanced metrics", "count", len(job.EnhancedMetrics), "namespace", svc.Namespace)
+	enhancedMetricData, err := enhancedMetricsService.GetMetrics(
```
jeschkies, line 99: So we don't need a processor anymore?

pkg/job/scrape.go
```diff
@@ -38,7 +39,27 @@ func ScrapeAwsData(
...
+		// - there was no error initializing the enhanced metrics processor previously
+		//
+		// if the initialization fails, we log the error and continue without enhanced metrics
+		if discoveryJob.HasEnhancedMetrics() &&
+			enhancedMetricsService == nil &&
+			enhancedMetricsServiceErr == nil {
+
+			enhancedMetricsService, enhancedMetricsServiceErr = enhancedmetrics.NewService(factory)
+			if enhancedMetricsServiceErr != nil {
+				logger.Warn("Couldn't initialize enhanced metrics processor", "err", enhancedMetricsServiceErr)
```
jeschkies:
```suggestion
				logger.Warn("Couldn't initialize enhanced metrics service", "err", enhancedMetricsServiceErr)
```

jeschkies (changes requested):
Thanks. We are getting there :slightly_smiling_face: I'd be easier if you'd only propose two services for the start.

docs/configuration.md
andriikushch: There is no implementation in this PR regarding this comment. This information came from [doc](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/elasticache@v1.51.8/types#CacheCluster:~:text=//%20The%20number%20of%20cache%20nodes%20in%20the%20cluster.%0A%09//%0A%09//%20For%20clusters%20running%20Valkey%20or%20Redis%20OSS%2C%20this%20value%20must%20be%201.%20For%20clusters%0A%09//%20running%20Memcached%2C%20this%20value%20must%20be%20between%201%20and%2040.).

pkg/internal/enhancedmetrics/service/rds/service.go
andriikushch: This value is returned in `GiB`. However, I think we can report it as bytes, as it is a more common representation of the memory in the metrics.

pkg/job/scrape.go
andriikushch: good catch

pkg/job/discovery.go
andriikushch: Correct. I removed the processor and replaced it with a service that encapsulates (extracts) factory type casting, some resource filtering, and interaction with the registry.  I would suggest keeping it for now; in the future, we might replace it with a direct registry usage.

pkg/internal/enhancedmetrics/service/rds/service.go
andriikushch: `buildAllocatedStorageMetric` doesn't require receiver `(s *RDS)`, however, this function has a high cohesion with the service.

andriikushch: However, it might not be a bad idea. Let's do it, as it is already scoped to the small package.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: Done

pkg/internal/enhancedmetrics/service/elasticache/service.go
andriikushch: Correct, we call the function that knows how to build a specific metric from the data that we received from the AWS API. It is defined in the `svc.supportedMetrics` in the service object.

pkg/internal/enhancedmetrics/service/elasticache/service.go
andriikushch: Changed.

pkg/internal/enhancedmetrics/service/dynamodb/service_test.go
andriikushch: Already changed

pkg/internal/enhancedmetrics/service/lambda/client.go
andriikushch: Correct, it is a signature from the AWS SDK.

pkg/internal/enhancedmetrics/service/lambda/service.go
andriikushch: Already changed

pkg/internal/enhancedmetrics/service/rds/service.go
andriikushch: Changed

pkg/internal/enhancedmetrics/service/elasticache/service.go
andriikushch: Renamed

pkg/internal/enhancedmetrics/service/lambda/service.go
andriikushch: In my opinion, it is not strictly necessary; however, if it will improve readability, let's change it. 

Speaking of the method name, I would prefer something like `buildEnhancedMetric`.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: With a current implementation, the registry returns a stateless service instance. It has the necessary parameters to initialize/create the client during the scrape (role and region). Also, for testing purposes, we need to have a way to use clients' mocks there. Therefore, I decided to use `buildClientFunc` here now so we can control which type of clients are produced at runtime. Result of the `NewXXXService` acts as a "template" when the registry returns an instance from the `Instance`.

pkg/internal/enhancedmetrics/service.go
andriikushch: I am fine with that. Let's do it.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: You are correct, it is not needed now here. It was used in the previous iterations to filter resources. 

I will remove it.

pkg/internal/enhancedmetrics/service/dynamodb/service.go
andriikushch: Good point. 

I think adding this type of validation to the job is a nice idea.

At the same time, I think the service should be resistant to incorrect usage (at least not cause panics); therefore, some level of validation should be in place, at least to avoid nil-pointer errors when we resolve the metric builder by metric name. Also, because the registry is a mutable global variable, its state can change between service job validation and specific service usage. Therefore, as an additional level of protection and to ensure correctness, I suggest adding some unsupported namespaces/metrics filtering; however, I do not think we need to do it at the specific service layer; it can be done one level above.

pkg/internal/enhancedmetrics/service/rds/client.go
andriikushch: Nice catch 👍

pkg/internal/enhancedmetrics/service/rds/service.go
andriikushch: yes 👍 

- https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1795#discussion_r2675839374

pkg/internal/enhancedmetrics/service/dynamodb/client.go
andriikushch: yes

pkg/config/config.go
```diff
@@ -238,7 +244,7 @@ func (j *Job) validateDiscoveryJob(logger *slog.Logger, jobIdx int) error {
 	if len(j.Regions) == 0 {
 		return fmt.Errorf("Discovery job [%s/%d]: Regions should not be empty", j.Type, jobIdx)
 	}
-	if len(j.Metrics) == 0 {
+	if len(j.Metrics) == 0 && len(j.EnhancedMetrics) == 0 {
 		return fmt.Errorf("Discovery job [%s/%d]: Metrics should not be empty", j.Type, jobIdx)
```
tristanburgess:
```suggestion
		return fmt.Errorf("Discovery job [%s/%d]: Metrics and EnhancedMetrics should not both be empty", j.Type, jobIdx)
```

pkg/internal/enhancedmetrics/service.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+
+	// filter out metrics that are not supported by the service
+	var filteredMetrics []*model.EnhancedMetricConfig
+	for _, metric := range metrics {
+		if svc.IsMetricSupported(metric.Name) {
+			filteredMetrics = append(filteredMetrics, metric)
+		} else {
+			logger.Debug("Skipping unsupported enhanced metric for service",
```
tristanburgess: Should this be a warning? If a user configures an enhanced metric incorrectly, like say a name typo etc, shouldn't they be made expressly aware of that?

docs/configuration.md
```diff
@@ -118,6 +118,15 @@ dimensionNameRequirements:
 # useful when cloudwatch metrics might not be present or when using info metrics to understand where your resources exist
 [ includeContextOnInfoMetrics: <boolean> ]
 
+# (optional) This is an experimental feature that can be used to enable enhanced metrics for specific services within this discovery job. It might be subject to changes in future releases.
```
tristanburgess, line 121: I wonder if this is the best place to put the supported enhanced metrics and their descriptions, also the list is duplicated in the `enhanced_metrics_config` section and will have to be separately updated it looks like. Maybe there should be a separate `enhanced_metrics.md` doc that has the list and these comments can point to it?

pkg/config/config.go
andriikushch: Good point 👍

pkg/internal/enhancedmetrics/service.go
andriikushch: I think it is fine to make it a warning; however, this should not happen during normal operations, because we have a check at the job level.

docs/configuration.md
andriikushch: I will remove a duplicate and keep a detailed description in the `enhanced_metrics_config` section below. 

I think that it is a little bit easier to find when everything is on one page, and the current list of the new metrics is relatively small.

Also, in my opinion, `enhanced_metrics.md` is a good idea for the next PR, where we can also add a guide on how to develop new/custom service components to extend the list of supported enhanced metrics.

docs/configuration.md
```diff
@@ -414,4 +418,20 @@ This is an example of the `dimensions_config` block:
...
+Currently supported enhanced metrics are:
+
+- AWS/Lambda (Timeout) - The maximum execution duration permitted for the function before termination.
+- AWS/DynamoDB (ItemCount) - The count of items in the table, updated approximately every six hours; may not reflect recent changes.
+- AWS/RDS (AllocatedStorage) - The storage capacity in bytes allocated for the DB instance.
+- AWS/ElastiCache (NumCacheNodes) - The count of cache nodes in the cluster; must be 1 for Valkey or Redis OSS clusters, or between 1 and 40 for Memcached clusters.
+
+```yaml
+enhancedMetrics:
+    - name: ItemCount
```
jeschkies, line 436: It's funny. I still wouldn't know what the possible values would be. Is it `Timeout`, `ItemCount`, `NumCacheNodes`? As I understand the enhanced metrics are nested under a type, aka namespace right? What happens when I put `NumCacheNodes` under `EC2`> ? Also, what is the motivation for using `name` here? Is this because we might add parameters in the future?

pkg/exporter_test.go
```diff
@@ -0,0 +1,255 @@
+// Copyright 2026 The Prometheus Authors
```
jeschkies, line 1: Could you rebase onto master? `master` should have this [file](https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/master/pkg/exporter_test.go) so I'm not sure why there is this diff

pkg/internal/enhancedmetrics/registry_test.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+func newMockService(namespace string) *registryMockMetricsServiceWrapper {
+	return &registryMockMetricsServiceWrapper{
+		namespace: namespace,
+		service: &registryMockMetricsService{
+			namespace: namespace,
+		},
+	}
+}
```
jeschkies: The tests never set `getMetricsFunc` nor `isMetricSupportedFunc` or am I missing something? I think you can replace `registryMockMetricsService` with `registryMockMetricsServiceWrapper` and simplify the tests.

pkg/internal/enhancedmetrics/registry_test.go
```diff
@@ -0,0 +1,311 @@
...
+
+func TestRegistry_Register(t *testing.T) {
```
jeschkies, line 43: This is just a nit but the tests could be table driven.

pkg/internal/enhancedmetrics/service.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/prometheus-community/yet-another-cloudwatch-exporter/pkg/model"
+)
+
+// MetricsServiceRegistry defines an interface to get enhanced metrics services by namespace
+type MetricsServiceRegistry interface {
+	GetEnhancedMetricsService(namespace string) (service.EnhancedMetricsService, error)
+}
+
+// Service is responsible for getting enhanced metrics using appropriate services.
+type Service struct {
```
jeschkies, line 32: The name is a little confusing. What is the difference between `Service` and `service.EnhancedMetricsService`?

pkg/internal/enhancedmetrics/service/rds/service.go
jeschkies: I've once read a doc by the original S3 author and they said that bytes should be used everywhere to avoid confusion 🙂

pkg/internal/enhancedmetrics/service.go
andriikushch: In this implementation:

- `enhancedmetrics.Service` is a "top-level" object that wraps the config provider and the registry. It hides the underlying logic for orchestrating metric extraction.
- `service.EnhancedMetricsService` - is an interface that should be implemented by the services that are getting specific enhanced metrics.

pkg/internal/enhancedmetrics/registry_test.go
andriikushch: changed

pkg/internal/enhancedmetrics/registry_test.go
andriikushch: Valid point 👍 I have removed `registryMockMetricsService`.

pkg/exporter_test.go
andriikushch: I have done a sync; I can't see this file in the diff anymore, but it is not marked as "Outdated" here.

docs/configuration.md
andriikushch: > Is it Timeout, ItemCount, NumCacheNodes?
Currently, there are 4 supported metrics:

`Timeout` for the AWS/Lambda
`ItemCount` - AWS/DynamoDB
`AllocatedStorage` - AWS/RDS
`NumCacheNodes` -  AWS/ElastiCache


> What happens when I put NumCacheNodes under EC2

In this case, an error will be returned, as there are no enhanced metrics for EC2.

- https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1795/changes#diff-90ef816253f3bc5b1c54c596f34fe58c982598079af347a6bda6daa60d615e73R48-R51

If the namespaces have enhanced metrics, unsupported metrics will be filtered, and a warning will be logged. 

- https://github.com/prometheus-community/yet-another-cloudwatch-exporter/pull/1795/changes#diff-90ef816253f3bc5b1c54c596f34fe58c982598079af347a6bda6daa60d615e73R68-R80

> Also, what is the motivation for using name here? Is this because we might add parameters in the future?

Correct, it gives an option to extend the config in the future.

tristanburgess (approved):

PR #1858, opened 2026-05-06
BE-1454 - Split CloudWatch rate limits per account region

omerkarj:
## Summary
- Split CloudWatch API rate limiters into independent `(account_id, region)` buckets while preserving the existing per-API limit values.
- Pass the resolved AWS account ID into CloudWatch client construction so discovery, custom namespace, static, `ListMetrics`, `GetMetricData`, and `GetMetricStatistics` use the correct bucket.
- Add `account_id` to rate-limit telemetry and cover independent account/region buckets with a regression test.

## Test plan
- `go test ./pkg/clients/cloudwatch`
- `go test ./pkg/clients/... ./pkg/job/... ./pkg/promutil`
- `go test ./...`

Made with [Cursor](https://cursor.com)


omerkarj:
Opened against the upstream repo by mistake; creating the intended PR in the groundcover-com fork.

PR #1889, opened 2026-06-30
Add support for reading a config from an env variable

atheken:
In some contexts (such as AWS ECS Fargate), it is somewhat involved to provide a config via a file on disk. This involves setting up sidecars/rendering, or placing files in external stores (such as S3).

This PR adds the ability to specify a config via a user-provided env var using a [similar convention to the standard OtelCollector](https://opentelemetry.io/docs/collector/configuration/#location).

Ultimately, this allows a user of YACE to use the official docker image in ECS Fargate, without needing to extend the container, or place files in some shared storage.


SuperQ (changes requested):
Sorry, but this is not something we intend to support as it doesn't fit with the Prometheus ecosystem.

PR #1912, opened 2026-07-24
feat: add bedrock-mantle namespace

superomel:
This change registers the AWS/BedrockMantle CloudWatch namespace and enables discovery of Mantle project resources

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Only the latest version gets security updates. We won't support older versions.
   * `AWS/AutoScaling` - Auto Scaling Group
   * `AWS/Backup` - Backup
   * `AWS/Bedrock` - GenerativeAI
+  * `AWS/BedrockMantle` - GenerativeAI OpenAI compatible
   * `AWS/Billing` - Billing
   * `AWS/Cassandra` - Cassandra
   * `AWS/CertificateManager` - Certificate Manager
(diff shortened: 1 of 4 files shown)


p-paul:
Interested in this one too — we're consuming `AWS/BedrockMantle` today via a `customNamespace` job, which works but gives up tag discovery:

```yaml
customNamespace:
  - name: bedrock-mantle
    namespace: AWS/BedrockMantle
    regions: [eu-west-1]
    dimensionNameRequirements: [Model, Project]
```

Two notes from running it that may be worth folding into this PR:

- **Both dimensions are mandatory.** These metrics are published with `Model` *and* `Project`, and CloudWatch matches on the exact dimension set — querying `Model` alone returns nothing, and returns it as zero rather than as an error, so it reads as "no usage" instead of "wrong query".
- **The metric names differ from `AWS/Bedrock`**: `InputTokens` / `OutputTokens` / `Inferences` here, versus `InputTokenCount` / `OutputTokenCount` / `Invocations` there. Not interchangeable, and an easy thing to copy wrong.

Also worth flagging for anyone landing here from the metric name: a custom namespace of `AWS/BedrockMantle` emits `aws_bedrockmantle_*`, not `aws_bedrock_mantle_*`. We had six dashboard panels silently returning nothing because of that.

Region is a constraint rather than a preference: `bedrock-mantle.<region>.api.aws` only resolves in `eu-west-1`.

PR #1913, opened 2026-07-31
Discover AWS/Bedrock application inference profiles for tag/name enrichment

AdilKh4n:
## Summary

`AWS/Bedrock` has no `ResourceFilters` entry in `pkg/config/services.go`, so the tagging client's `resourcegroupstaggingapi`-based discovery (`pkg/clients/tagging/client.go`) never runs for it (`if len(svc.ResourceFilters) > 0`). As a result, `AWS/Bedrock` metrics are always emitted with:

- Empty `tag_*` labels, regardless of `exportedTagsOnMetrics` configuration
- A `name="global"` label, since no resource is ever associated to the metric

This is easy to reproduce: configure `exportedTagsOnMetrics.AWS/Bedrock` with any tag key, scrape `AWS/Bedrock` metrics, and the resulting `tag_*` label value is always `""`.

Unlike most namespaces, Bedrock application inference profile tags aren't reliably obtainable through the tagging API in a way that also carries the profile's human-readable name (`InferenceProfileName`), which isn't part of the ARN. So instead of adding `ResourceFilters`, this PR adds a dedicated `ServiceFilter` (see the existing `AWS/StorageGateway` entry for precedent) that:

- Calls `bedrock:ListInferenceProfiles` directly to discover inference profiles
- Calls `bedrock:ListTagsForResource` per profile to fetch tags (best-effort: `SYSTEM_DEFINED` profiles are AWS-owned and don't support tagging, so a failed call is treated as "no tags" rather than aborting discovery)
- Sets the resource's `ARN` field to `"<profile-id>/<profile-name>"` instead of the real ARN (same trick `AWS/StorageGateway` uses with `"<gateway-id>/<gateway-name>"`), since CloudWatch's `ModelId` dimension for these metrics is the bare profile ID, not the full ARN. This lets the `ModelId` dimension regex match the id prefix while the profile's name still ends up in the metric's `name` label instead of `"global"`.

## Changes

- `pkg/config/services.go`: drop `ResourceFilters` from `AWS/Bedrock`, update the `DimensionRegexps` entry to match the new `<id>/<name>` scheme
- `pkg/clients/tagging/filters.go`: add a `ServiceFilter` for `AWS/Bedrock`
- `pkg/clients/tagging/adapters.go`, `client.go`: wire up a `bedrock` client adapter, following the existing closure-based adapter pattern
- `pkg/clients/factory.go`: construct the `*bedrock.Client`
- `pkg/promutil/prometheus.go`: add `BedrockAPICounter` scrape metric
- `pkg/job/maxdimassociator/associator_bedrock_test.go`: new associator test covering the match/skip cases
- `README.md`: document the new `bedrock:ListInferenceProfiles` / `bedrock:ListTagsForResource` IAM permissions
- `go.mod`/`go.sum`: add `github.com/aws/aws-sdk-go-v2/service/bedrock` (minor patch bumps to `aws-sdk-go-v2` core and `smithy-go` came along via `go mod tidy`)

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (all packages pass)
- [x] Verified against a real account: the previously-broken `tag_app` label (added via `exportedTagsOnMetrics`) and `name` label are both populated correctly once this discovery path runs

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,8 @@ As a quick start, the following IAM policy can be used to grant the all permissi
         "apigateway:GET",
         "aps:ListWorkspaces",
         "autoscaling:DescribeAutoScalingGroups",
+        "bedrock:ListInferenceProfiles",
+        "bedrock:ListTagsForResource",
         "dms:DescribeReplicationInstances",
         "dms:DescribeReplicationTasks",
         "ec2:DescribeTransitGatewayAttachments",
@@ -209,6 +211,13 @@ This permission is required to discover resources for the AWS/ApiGateway namespa
 This permission is required to discover resources for the AWS/AutoScaling namespace
 ```json
 "autoscaling:DescribeAutoScalingGroups"
+```
+
+These permissions are required to discover resources (application inference profiles) and their
+tags for the AWS/Bedrock namespace
+```json
+"bedrock:ListInferenceProfiles",
+"bedrock:ListTagsForResource"
 ```
 
 These permissions are required to discover resources for the AWS/DMS namespace
--- a/go.mod
+++ b/go.mod
@@ -3,13 +3,14 @@ module github.com/prometheus-community/yet-another-cloudwatch-exporter
 go 1.25.0
 
 require (
-	github.com/aws/aws-sdk-go-v2 v1.42.1
+	github.com/aws/aws-sdk-go-v2 v1.43.3
 	github.com/aws/aws-sdk-go-v2/config v1.32.30
 	github.com/aws/aws-sdk-go-v2/credentials v1.19.29
 	github.com/aws/aws-sdk-go-v2/service/amp v1.45.2
 	github.com/aws/aws-sdk-go-v2/service/apigateway v1.41.1
 	github.com/aws/aws-sdk-go-v2/service/apigatewayv2 v1.36.1
 	github.com/aws/aws-sdk-go-v2/service/autoscaling v1.69.1
+	github.com/aws/aws-sdk-go-v2/service/bedrock v1.66.3
 	github.com/aws/aws-sdk-go-v2/service/cloudwatch v1.63.1
 	github.com/aws/aws-sdk-go-v2/service/databasemigrationservice v1.65.1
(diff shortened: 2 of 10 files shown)


p-paul:
Independently hit and reproduced this on v0.62.1 — confirming the diagnosis in the description, with the evidence in case a second data point helps.

We tag AWS Bedrock application inference profiles with `AICodeUser=<person>` to attribute Claude Code usage, and configured:

```yaml
discovery:
  exportedTagsOnMetrics:
    AWS/Bedrock:
      - AICodeUser
```

Every `aws_bedrock_*` series came back with `name="global"` and no `tag_*` label at all — 207 series over 90 days, not one tagged. Same exporter, same scrape, `AWS/FSx` and `AWS/NetworkELB` get `name=<ARN>` plus their full `tag_*` set, so tag export itself is working; `AWS/Bedrock` specifically cannot use it. That matches the `if len(svc.ResourceFilters) > 0` path you identified.

Two things that might save the next person some time:

- The failure is completely silent. `exportedTagsOnMetrics` accepts the namespace, validates fine, and produces empty labels rather than warning that the namespace has no `ResourceFilters`. It reads as a tagging/IAM problem, and the tags are verifiably present — `resourcegroupstaggingapi get-resources --resource-type-filters bedrock` returns every profile with its tags.
- There is no `aws_bedrock_info` metric either (no resources are discovered), so the usual `metric * on(name) group_left(tag_X) aws_<service>_info` workaround has nothing to join against.

On the approach: setting `ARN` to `<profile-id>/<profile-name>` is the part that makes this worth more than a plain `ResourceFilters` entry. `InferenceProfileName` is genuinely not in the ARN, so a `DimensionRegexps` patch alone could recover the tags but never the readable name — and the bare profile id (a 12-character opaque id) is exactly what makes these metrics useless on a dashboard.

We've worked around it by moving `AWS/Bedrock` to `prometheus/cloudwatch_exporter`, where `aws_tag_select.resource_id_dimension` makes the mapping configurable, and left everything else on YACE. We'd move it back if this lands — the `name` handling here gives a better result than the workaround does. Happy to test a build against a real account with tagged application inference profiles if that's useful.

Issue #1853, opened 2026-04-23
YACE using excessive iam:ListAccountAliases calls


caitlin-green-sp:
YACE very frequently uses iam:ListAccountAliases calls, to the point where it's eating up our rate-limits within AWS. Is there any way to reduce or not require this permission for YACE to function?  It says this permission is optional, but when we didn't include it yace still "works" without "iam:ListAccountAliases" it just calls IAM at least twice as many times if you remove the permission from the policy.

Additionally, I have seen this issue:
https://github.com/prometheus-community/yet-another-cloudwatch-exporter/issues/1541

We would like to know if there is any ways whatsoever to reduce the number of IAM calls or make this permission truly optional.

Issue #1541, opened 2024-10-14
[BUG] When "iam:ListAccountAliases" is not granted yace errors and will not function.


mikeq3:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### YACE version

0.61.2

### Current Behavior

When "iam:ListAccountAliases" is not granted yace errors and will not function.

### Expected Behavior

When "iam:ListAccountAliases" is not granted yace warns and continues to function.

### Anything else?

The change log states that IF "iam:ListAccountAliases" is granted then the account alias is used. It does not say that it has to be enabled. I thought from the wording that this is optional. I believe it should be optional.

carlduevel:
Are you sure the exporter stopped working? It issues a [WARN log statement for sure,](https://github.com/prometheus-community/yet-another-cloudwatch-exporter/blob/aa362379f2cf1f5208fbeb9909ced10bb170c674/pkg/job/scrape.go#L111) but I did not observe errors caused by this.

PR #1925, opened 2026-08-24
Add -disable-account-alias-lookup flag

NickAdolf:
Adds a CLI flag to skip the IAM ListAccountAliases API call during scrapes. When enabled, the account_alias label on metrics will be empty.

This is useful in environments where:
- IAM global endpoint access is restricted or undesirable. For us this is generating tons of undue firewall egress sessions and the call isn't needed in our config.
- The account alias is redundant with customTags configuration. In our case we don't need the alias as its already a custom tag due to dynamic config generation.

Related to #1853
Related to #1541

### DIFF
--- a/cmd/yace/main.go
+++ b/cmd/yace/main.go
@@ -54,17 +54,18 @@ const (
 )
 
 var (
-	addr                  string
-	configFile            string
-	logLevel              string
-	logFormat             string
-	fips                  bool
-	cloudwatchConcurrency config.CloudWatchConcurrencyConfig
-	tagConcurrency        int
-	scrapingInterval      int
-	metricsPerQuery       int
-	labelsSnakeCase       bool
-	profilingEnabled      bool
+	addr                      string
+	configFile                string
+	logLevel                  string
+	logFormat                  string
+	fips                      bool
+	disableAccountAliasLookup bool
+	cloudwatchConcurrency     config.CloudWatchConcurrencyConfig
+	tagConcurrency            int
+	scrapingInterval          int
+	metricsPerQuery           int
+	labelsSnakeCase           bool
+	profilingEnabled          bool
 
 	logger *slog.Logger
 )
@@ -138,6 +139,12 @@ func NewYACEApp() *cli.App {
 			Value:       false,
 			Usage:       "Use FIPS compliant AWS API endpoints",
 			Destination: &fips,
+		},
+		&cli.BoolFlag{
+			Name:        "disable-account-alias-lookup",
+			Value:       false,
+			Usage:       "Disable the IAM ListAccountAliases call. The account_alias label on metrics will be empty.",
+			Destination: &disableAccountAliasLookup,
 		},
 		&cli.IntFlag{
 			Name:        "cloudwatch-concurrency",
@@ -262,6 +269,7 @@ func startScraper(c *cli.Context) error {
 	cfg.TaggingAPIConcurrency = tagConcurrency
 	cfg.FeatureFlags = c.StringSlice(enableFeatureFlag)
 	cfg.FIPSEnabled = fips
+	cfg.DisableAccountAliasLookup = disableAccountAliasLookup
 	cfg.CloudwatchConcurrency = cloudwatchConcurrency
 	if err := cfg.Validate(); err != nil {
 		return fmt.Errorf("invalid runtime scrape configuration: %w", err)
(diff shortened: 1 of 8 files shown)


stevo-f3:
@andriikushch PTAL 🙏

cmd/yace/main.go
```diff
@@ -54,17 +54,18 @@ const (
...
-	cloudwatchConcurrency config.CloudWatchConcurrencyConfig
-	tagConcurrency        int
-	scrapingInterval      int
-	metricsPerQuery       int
-	labelsSnakeCase       bool
-	profilingEnabled      bool
+	addr                      string
+	configFile                string
+	logLevel                  string
+	logFormat                  string
```
andriikushch: Could you please run `go fmt ./cmd/yace`? (There is an extra space in this line)

pkg/job/scrape.go
```diff
@@ -35,6 +35,7 @@ func ScrapeAwsData(
 	metricsPerQuery int,
 	cloudwatchConcurrency cloudwatch.ConcurrencyConfig,
 	taggingAPIConcurrency int,
+	disableAccountAliasLookup bool,
```
andriikushch, line 39: I would consider adding some tests to cover the new functionality. What do you think?

pkg/config/config.go
```diff
@@ -47,13 +47,14 @@ type CloudWatchConcurrencyConfig struct {
...
-	FeatureFlags          []string
-	FIPSEnabled           bool
-	CloudwatchConcurrency CloudWatchConcurrencyConfig
+	ScrapeConfigFile          string
+	MetricsPerQuery           int
+	LabelsSnakeCase           bool
+	TaggingAPIConcurrency     int
+	FeatureFlags              []string
+	FIPSEnabled               bool
+	DisableAccountAliasLookup bool
```
andriikushch, line 56: nit: I think we could set it explicitly in the `DefaultConfig`, similar to `FIPSEnabled`.

pkg/job/scrape.go
```diff
@@ -140,9 +144,12 @@ func ScrapeAwsData(
...
-					accountAlias, err := factory.GetAccountClient(region, role).GetAccountAlias(ctx)
-					if err != nil {
-						jobLogger.Warn("Couldn't get account alias", "err", err)
+					var accountAlias string
+					if !disableAccountAliasLookup {
+						accountAlias, err = factory.GetAccountClient(region, role).GetAccountAlias(ctx)
+						if err != nil {
+							jobLogger.Warn("Couldn't get account alias", "err", err)
+						}
 					}
```
andriikushch: nit: That fetch-and-warn logic is copied and pasted a few times throughout the code. I am wondering if we could find a good abstraction/name so we don't have to repeat it. It shouldn't block the PR; in my opinion, we can change it in upcoming PRs.

pkg/job/scraper.go
```diff
@@ -84,11 +87,13 @@ func (s Scraper) Scrape(ctx context.Context) ([]model.TaggedResourceResult, []mo
 			a := Account{
 				ID: accountID,
 			}
-			accountAlias, err := client.GetAccountAlias(ctx)
-			if err != nil {
-				s.logger.Warn("Failed to get optional account alias from account", "err", err, "account_id", accountID)
-			} else {
-				a.Alias = accountAlias
+			if !s.disableAccountAliasLookup {
```
andriikushch: nit: because of the config name we are using double-negation here, we could try to avoid it.

andriikushch (commented):
Hi @NickAdolf 👋 

Thank you very much for the PR. 

Overall, it looks good to me. 👍 I left a few comments; please review them.

pkg/job/scrape.go
NickAdolf: I've added some based on the existing to provide coverage.

pkg/job/scrape.go
NickAdolf: I took a swing at collapsing this to keep things DRY and to address the double negative.

NickAdolf: Definitely touches a bit more than before, but seems to be the easiest to follow of the options I found.

pkg/job/scraper.go
NickAdolf: took a new approach here that has removed this.

andriikushch (approved):
Nice job @NickAdolf! 👍 

Thank you for the PR! I will approve it.

NickAdolf:
@andriikushch just for internal planning purposes, do we know when we expect the next release cut?

andriikushch:
HI @NickAdolf 👋 

> just for internal planning purposes, do we know when we expect the next release cut?

I plan to do it next week (we might want to add one more PR to the release, but it is in progress right now).

PR #1926, opened 2026-08-28
feat: Infer ARN from dimensions during discovery.

jeschkies:
This changes adds a few missing dimensions regexp to match more namespace to tagged resources. It also offers an optional fallback that infers the ARN and thus resource name from the metric dimensions in case a resources isn't tagged and couldn't be found using `GetTaggedResources`.

### DIFF
--- a/pkg/config/scrapeconf.go
+++ b/pkg/config/scrapeconf.go
@@ -466,6 +466,7 @@ func (c *ScrapeConf) toModelConfig() model.JobsConfig {
 		job.IncludeContextOnInfoMetrics = discoveryJob.IncludeContextOnInfoMetrics
 		job.DimensionsRegexps = svc.ToModelDimensionsRegexp()
 		job.EnhancedMetrics = svc.toModelEnhancedMetricsConfig(discoveryJob.EnhancedMetrics)
+		job.ArnFallback = svc.toArnFallback()
 
 		job.ExportedTagsOnMetrics = []string{}
 		if len(c.Discovery.ExportedTagsOnMetrics) > 0 {
--- a/pkg/config/scrapeconf.go
+++ b/pkg/config/scrapeconf.go
@@ -69,6 +69,7 @@ type Job struct {
 	RecentlyActiveOnly          bool              `yaml:"recentlyActiveOnly"`
 	IncludeContextOnInfoMetrics bool              `yaml:"includeContextOnInfoMetrics"`
 	EnhancedMetrics             []*EnhancedMetric `yaml:"enhancedMetrics"`
+	EnableArnFallback           bool              `yaml:"enabled_arn_fallback"`
 	JobLevelMetricFields        `yaml:",inline"`
 }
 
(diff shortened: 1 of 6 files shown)


pkg/config/scrapeconf.go
```diff
@@ -69,6 +69,7 @@ type Job struct {
 	RecentlyActiveOnly          bool              `yaml:"recentlyActiveOnly"`
 	IncludeContextOnInfoMetrics bool              `yaml:"includeContextOnInfoMetrics"`
 	EnhancedMetrics             []*EnhancedMetric `yaml:"enhancedMetrics"`
+	EnableArnFallback           bool              `yaml:"enabled_arn_fallback"`
```
bboreham, line 72: YAML name seems inconsistent - other ones in this struct are camelCase and have no underscores.

Also should it say "enabled" or "enabled" ?

pkg/config/services.go
```diff
@@ -21,6 +22,105 @@ import (
...
+// ArnFromDimensionsFunc builds a resource ARN directly from a ListMetrics result's dimensions,
+// without requiring the resource to have first been discovered through the Tagging API (GetResources).
+// dimensions is keyed by the same names used in this file's DimensionRegexps named capture groups
+// (e.g. "FunctionName", "TableName", "Directory_ID"), which matches the real AWS CloudWatch dimension
+// name for that namespace with spaces replaced by underscores.
+//
+// Returns false when the supplied dimensions don't contain enough information to build a valid ARN.
+//
+// These are best-effort, statically derived from documented AWS ARN formats and have not been
+// validated against live AWS API responses; verify against real output before relying on them.
```
bboreham, line 34: Do we expect the relevant audience is reading this code, in order to heed this warning?

Should it be repeated in documentation?

pkg/config/scrapeconf.go
tristanburgess: Hey Bryan yes this should be camelCase and should be `enable` instead of `enabled`, I will also think if the name can be improved to make it more clear what it's for without reading further code + docs.

pkg/config/services.go
tristanburgess: Thanks agreed, it should be repeated in `docs/configuration.md`. I'll make the changes

pkg/job/discovery.go
tristanburgess: Doesn't this condition hold for untagged resources? Therefore we never end up doing inference

pkg/job/discovery.go
```diff
@@ -201,18 +205,24 @@ func getFilteredMetricDatas(
 
 		resource := matchedResource
 		if resource == nil {
```
tristanburgess: This case I believe was more intended for there being tagged resources but the dimension set in the metric itself did not contain a key/value pair that would be able to tie it to an individual resource (hence arn := "global")

pkg/job/discovery.go
tristanburgess: FWIW https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html does return entries for "resources that were previously tagged" but are now untagged, so I think we need both cases covered by inference.

PR #1932, opened 2026-09-14
BE-3208 - Discover untagged API Gateway APIs for AWS/ApiGateway metrics

nir-sagi:
## Why

`AWS/ApiGateway` discovery relied on the Resource Groups Tagging API alone, so untagged REST/HTTP APIs were never discovered. Once at least one API *is* discovered, the associator drops every dimensioned metric that matches no resource, so `Latency`/`4XXError`/`5XXError` never arrive and `Count` shows only the account rollup. Seen on the Yalo POC (21 REST APIs, 4 series).

## What

- `AWS/ApiGateway` `ServiceFilter` gains a `ResourceFunc` (v1 and v2 clients, parity test requires both) that emits every REST and HTTP/WebSocket API, Id-keyed exactly like the tagging API returns them.
- The existing `FilterFunc` is unchanged in behavior: it renames REST ARNs to `ApiName` and, since each API matches once, drops the untagged duplicate of any API the tagging service already returned with its tags. Tagged resources keep priority and their tags.
- `searchTags` jobs still exclude untagged APIs (`FilterThroughTags`).
- The two API fetches are shared through one `listApiGateways` helper; the ARN builder is a pure function with a test.

Supersedes groundcover-private#26765 (in-repo lister). Follow-up: tag a release and bump the `replace` in groundcover-private `go.mod`.

## Test

`go test ./pkg/...` — 569 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nir-sagi:
Opened against the wrong repo by mistake, sorry for the noise.

